PR #8, opened 2018-07-26
Customization and use of two designs (widget input) with a boolean preference

cathirv:
This PR is pending another (https://github.com/Kommunicate-io/Kommunicate-Android-Chat-SDK/pull/7), so the first commit would be --> 
Add new design widget_input_layout.xml -->
b4516d1


mukeshthawani:
Thanks for contributing @cathirv. Team is reviewing it and we’ll make sure it gets merged in the next couple of days.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #59, opened 2019-08-14
feedback+custom title

shubhamtewari:

PR #60, opened 2019-08-21
Set conversation title to channelName and reverse dependencies

reytum:
1) The conversation title has been replaced by channel.getName() from the name of the assignee.
2) The dependencies have been reversed for kommunicate and kommunicateui


kommunicate/src/main/java/io/kommunicate/KmConversationHelper.java
```diff
@@ -263,7 +268,7 @@ private static KMStartChatHandler getStartChatHandler(final boolean isSkipChatLi
             public void onSuccess(Channel channel, Context context) {
                 try {
                     if (resultReceiver != null) {
-                        resultReceiver.send(LeadCollectionActivity.PRECHAT_RESULT_CODE, null);
+                        resultReceiver.send(KmConstants.PRECHAT_RESULT_CODE, null);
```
devashishmamgain: It is considered a bad practice to keep all constants in a single file, constants should be placed where they belong to avoid any dependencies between different module or packages of code. https://github.com/Kommunicate-io/Kommunicate-Android-Chat-SDK/pull/59#pullrequestreview-278956711

kommunicate/src/main/java/io/kommunicate/async/KmConversationCreateTask.java
```diff
@@ -0,0 +1,73 @@
...
+public class KmConversationCreateTask extends AsyncTask<Void, Void, ChannelFeedApiResponse> {
+    private WeakReference<Context> context;
+    private ChannelService channelService;
+    private ChannelInfo channelInfo;
+    private KmStartConversationHandler startConversationHandler;
+    private KMStartChatHandler startChatHandler;
+
+
+    public KmConversationCreateTask(Context context, ChannelInfo channelInfo, KmStartConversationHandler startConversationHandler) {
+        this.context = new WeakReference<>(context);
```
devashishmamgain: We can create another constructor with common elements and then call this(..,..,..) from these 2

kommunicate/src/main/java/io/kommunicate/Kommunicate.java
```diff
@@ -513,4 +518,8 @@ private static String generateUserId() {
         }
         return text.toString();
     }
+
+    public static void getAwayMessage(Context context, Integer groupId, KmAwayMessageHandler handler) {
```
devashishmamgain: This method is not returning anything but naming with get makes it looks like it is going to return something. Also static method running something in async, is that correct way?

reytum: That method is will be returning the message in the Callback KmAwayMessageHandler.
It is just a wrapper that executes an async task. It is good practice as we do not have to implement the async task everywhere, just calling the method works. If android stops support for async task, then we can directly change the logic inside the method and the usages won't be effected.

kommunicate/src/main/java/io/kommunicate/KmConversationHelper.java
reytum: I guess its OK to keep constants shared across different classes in a single file. Since these constants are being referred to in a dependent module, I have placed them in a single file.

kommunicate/src/main/java/io/kommunicate/async/KmConversationCreateTask.java
reytum: Done. Actually, this was done as the second constructor is deprecated and will be removed after some time.

kommunicate/src/main/java/io/kommunicate/Kommunicate.java
devashishmamgain: Naming convention in above example is not 'get', its 'on'. All get in above example are returning object

kommunicate/src/main/java/io/kommunicate/async/KmConversationCreateTask.java
```diff
@@ -0,0 +1,73 @@
...
+        if (channelFeedApiResponse != null) {
+            if (channelFeedApiResponse.isSuccess()) {
+                if (startConversationHandler != null) {
+                    startConversationHandler.onSuccess(channelService.getChannel(channelFeedApiResponse.getResponse()), context.get());
+                }
+                if (startChatHandler != null) {
+                    startChatHandler.onSuccess(channelService.getChannel(channelFeedApiResponse.getResponse()), context.get());
+                }
+            } else {
+                if (startConversationHandler != null) {
```
devashishmamgain: @reytum this looks same to the code block in another else. is it possible to put it in some variable then have only 1 common code. 
  if (startConversationHandler != null) {
                startConversationHandler.onFailure(VARIABLE, context.get());
            }
            if (startChatHandler != null) {
                startChatHandler.onFailure(null, context.get());
            }

kommunicateui/src/main/res/values/strings.xml
```diff
@@ -115,6 +113,8 @@
     <string name="please_wait_creating_group_of_two">Please wait...</string>
     <string name="sort_contacts_by_lastSeenAt">Sort by lastSeen</string>
     <string name="sort_contacts_by_recent_chats">Sort by recent chats</string>
+    <string name="logout_info_text">"Logging out, please wait..."</string>
```
devashishmamgain: @reytum some places we have .. and some places it is ... 
is there any specific convention on this?

kommunicate/src/main/java/io/kommunicate/KmConversationHelper.java
reytum: Earlier these constants were defined in their respective classes. But now the dependencies have been reversed and thus those constants cannot be accessed from the classes that are in kommunicate module. If we move these to original classes then we will have to declare these constants again in the class that is accessing them which is similar to declaring them in a single file.

kommunicate/src/main/java/io/kommunicate/Kommunicate.java
reytum: Done. Changed it to loadAwayMessage

kommunicateui/src/main/res/values/strings.xml
reytum: We usually use 3 dots for loading texts. 2 dots might be old code.

kommunicate/src/main/java/io/kommunicate/async/KmConversationCreateTask.java
reytum: Here we need to trigger both the code blocks based on whether the response is null or the response is error. Did you mean that we need to move the complete failure code block to another method and then call the method in both the places?

kommunicateui/src/main/res/values/strings.xml
devashishmamgain: @shubhamtewari saw you were using 2 dots in a recent commit, can you check whats the standard way based on Google guidelines?

kommunicate/src/main/java/io/kommunicate/async/KmConversationCreateTask.java
devashishmamgain: yes

kommunicate/src/main/java/io/kommunicate/callbacks/KmStartConversationHandler.java
```diff
@@ -1,6 +1,13 @@
...
 
-import com.applozic.mobicomkit.uiwidgets.async.AlChannelCreateAsyncTask;
 
-public interface KmStartConversationHandler extends AlChannelCreateAsyncTask.TaskListenerInterface {
+import android.content.Context;
+
+import com.applozic.mobicomkit.feed.ChannelFeedApiResponse;
+import com.applozic.mobicommons.people.channel.Channel;
+
+public interface KmStartConversationHandler {
```
devashishmamgain: @reytum is it similar to KMStartChatHandler? can both use the same handler or it has to be different?

kommunicate/src/main/java/io/kommunicate/services/KmService.java
```diff
@@ -104,12 +106,12 @@ public static Contact getAssigneeContact(Channel channel, BaseContactService con
             String conversationAssignee = null;
             String conversationTitle = null;
 
-            if (metadataMap.containsKey(KommunicateUI.CONVERSATION_ASSIGNEE)) {
-                conversationAssignee = metadataMap.get(KommunicateUI.CONVERSATION_ASSIGNEE);
+            if (metadataMap.containsKey(KmConstants.CONVERSATION_ASSIGNEE)) {
```
devashishmamgain: @reytum is this and below code same as present in getSupportGroupContact ? can we move it to common function and reuse that.

kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/conversation/fragment/MobiComConversationFragment.java
```diff
@@ -4396,17 +4395,17 @@ public void processTakeOverFromBot(Context context, final Channel channel) {
...
+        KmService.removeMembersFromConversation(context, channel.getKey(), botIds, new KmRemoveMemberCallback() {
             @Override
-            public void onSuccess(String response, int index, Context context) {
+            public void onSuccess(String response, int index) {
                 GroupInfoUpdate groupInfoUpdate = new GroupInfoUpdate(channel);
                 Map<String, String> metadata = channel.getMetadata();
                 if (metadata == null) {
                     metadata = new HashMap<>();
                 }
                 metadata.put("CONVERSATION_ASSIGNEE", loggedInUserId);
```
devashishmamgain: @reytum use constants

kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/kommunicate/utils/KmHelper.java
```diff
@@ -70,7 +75,7 @@ public void onSuccess(Object message) {
                 @Override
                 public void onFailure(Object error) {
                     dialog.dismiss();
-                    Toast.makeText(context, context.getString(R.string.unable_to_create_conversation) + " : " + error, Toast.LENGTH_SHORT).show();
+                    Toast.makeText(context, context.getString(io.kommunicate.R.string.unable_to_create_conversation) + " : " + error, Toast.LENGTH_SHORT).show();
```
devashishmamgain: @reytum is io.kommunicate.R intentional here?

kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/kommunicate/utils/KmHelper.java
```diff
@@ -82,23 +87,26 @@ public void onFailure(Object error) {
...
-                @Override
-                public void onFailure(Object error) {
-                    dialog.dismiss();
-                    Toast.makeText(context, context.getString(R.string.unable_to_create_conversation) + " : " + error, Toast.LENGTH_SHORT).show();
-                }
-            });
+                        @Override
+                        public void onFailure(Object error) {
+                            dialog.dismiss();
+                            Toast.makeText(context, context.getString(io.kommunicate.R.string.unable_to_create_conversation) + " : " + error, Toast.LENGTH_SHORT).show();
```
devashishmamgain: @reytum review if this is intentional here io.kommunicate.R, also we can remove space before :

kommunicate/src/main/java/io/kommunicate/callbacks/KmStartConversationHandler.java
reytum: The KmStartChatHandler is deprecated and will be removed. Both are different and used by different codes.

kommunicate/src/main/java/io/kommunicate/services/KmService.java
```diff
@@ -32,7 +34,9 @@
     private Context context;
     private KmClientService clientService;
     private KmAutoSuggestionDatabase autoSuggestionDatabase;
-
+    public static final String KM_SKIP_BOT = "skipBot";
+    public static final String KM_NO_ALERT = "NO_ALERT";
```
devashishmamgain: @reytum this must be present in Applozic sdk, or possibly moving these to Applozic sdk might be a better place as it is related to notifications and not necessarily kommunicate.

kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/conversation/fragment/MobiComConversationFragment.java
```diff
@@ -4396,29 +4395,29 @@ public void processTakeOverFromBot(Context context, final Channel channel) {
...
+                        message.setMessage(Utils.getString(context, R.string.km_assign_to_message) + new ContactDatabase(context).getContactById(loggedInUserId).getDisplayName());
 
                         Map<String, String> metadata = new HashMap<>();
-                        metadata.put("skipBot", "true");
-                        metadata.put("KM_ASSIGN", MobiComUserPreference.getInstance(context).getUserId());
-                        metadata.put("NO_ALERT", "true");
-                        metadata.put("BADGE_COUNT", "false");
-                        metadata.put("category", "ARCHIVE");
+                        metadata.put(KmService.KM_SKIP_BOT, Message.GroupMessageMetaData.TRUE.getValue());
+                        metadata.put(Message.KM_ASSIGN, MobiComUserPreference.getInstance(context).getUserId());
```
devashishmamgain: @reytum Assignment is a property of group(Channel), groups are assigned to agent, messages are not assigned to agent, do you think Channel will be a better place to put this constant?
check the same with other properties as well.

kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/conversation/richmessaging/payment/PaymentActivity.java
```diff
@@ -49,7 +50,7 @@ protected void onCreate(Bundle savedInstanceState) {
         txnData = new HashMap<>();
         setWebViewClient();
 
-        String helpCenterUrl = getIntent().getStringExtra(KommunicateUI.KM_HELPCENTER_URL);
+        String helpCenterUrl = getIntent().getStringExtra(KmConstants.KM_HELPCENTER_URL);
```
devashishmamgain: @reytum how is helpcenter related to Payment? (i see that this is an existing code, not related to this issue, but still curious to know, why it is placed in Payment?)

kommunicate/src/main/java/io/kommunicate/async/KmConversationCreateTask.java
reytum: Done

kommunicate/src/main/java/io/kommunicate/services/KmService.java
reytum: Done

kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/conversation/fragment/MobiComConversationFragment.java
reytum: Done

kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/kommunicate/utils/KmHelper.java
reytum: Done

kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/kommunicate/utils/KmHelper.java
reytum: Done

kommunicate/src/main/java/io/kommunicate/services/KmService.java
reytum: Yes, will move it to applozic SDK, but right now we need some constants for reference as applozic SDK is imported via gradle.

kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/conversation/fragment/MobiComConversationFragment.java
reytum: KM_ASSIGN and KM_STATUS are properties of messages as they are message level attributes. We check the same on Messages

kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/conversation/richmessaging/payment/PaymentActivity.java
reytum: During hotel booking bot development, this Activity was named as PaymentActivity, Since it does the web view reload work, we used it for all the web view related work later. Will have to rename it .

reytum:
> clarification on ... vs .. is remaining. pls check, rest all are cleared.

I have not been able to find anything related to it. All apps I have seen uses ... for loading texts as it is an ellipsis character. Google uses the same for all its apps like "Syncing mails..." for gmail etc.

devashishmamgain:
> > clarification on ... vs .. is remaining. pls check, rest all are cleared.
> 
> I have not been able to find anything related to it. All apps I have seen uses ... for loading texts as it is an ellipsis character. Google uses the same for all its apps like "Syncing mails..." for gmail etc.

@shubhamtewari is using 2 dots in the code and you are using 3 dots.
example:
https://github.com/Kommunicate-io/Kommunicate-Android-Chat-SDK/pull/59/files
    <string name="add_a_comment">Add a comment..</string>
    <string name="payment_successful">"Payment Successful.."</string>

and 
this PR
    <string name="logout_info_text">"Logging out, please wait..."</string>


If 3 dots is correct then please add review comment on https://github.com/Kommunicate-io/Kommunicate-Android-Chat-SDK/pull/59/files and get that corrected to 3 dots.

PR #61, opened 2019-09-11
add method updateChatContext

shubhamtewari:


kommunicate/src/main/java/io/kommunicate/Kommunicate.java
```diff
@@ -30,6 +31,11 @@
 import com.applozic.mobicommons.commons.core.utils.Utils;
 import com.applozic.mobicommons.json.GsonUtils;
 import com.applozic.mobicommons.people.channel.Channel;
+import com.google.gson.Gson;
+import com.google.gson.JsonObject;
+import com.google.gson.reflect.TypeToken;
+
+import org.json.JSONObject;
```
devashishmamgain: @shubhamtewari is this getting used?

devashishmamgain: @reytum is we are using Gson then do we need to use org.json.JSONObject as well?

devashishmamgain:
@shubhamtewari what is expected from this method:
lets say if there is already one key value pair stored as:

sessionId: abc

and new key value pair comes as:
language: spanish

then what is the expected final value:

session: abc
language: spanish

or just
language: spanish

kommunicate/src/main/java/io/kommunicate/Kommunicate.java
shubhamtewari: we dont require the import, removed it

shubhamtewari:
> @shubhamtewari what is expected from this method:
> lets say if there is already one key value pair stored as:
> 
> sessionId: abc
> 
> and new key value pair comes as:
> language: spanish
> 
> then what is the expected final value:
> 
> session: abc
> language: spanish
> 
> or just
> language: spanish

it the data previously stored in the KmChatContext key will be updated
session: abc
language: spanish

PR #75, opened 2019-10-21
add icon and alert dialog

shubhamtewari:
improvements for the new image upload flow


kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/conversation/activity/ConversationActivity.java
```diff
@@ -592,10 +593,41 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
...
+
+                                @Override
+                                public void onClickPositiveButton(DialogInterface dialog, int id) {
+                                    for (int i = 0; i < KmAttachmentsController.NO_OF_MULTI_SELECTIONS_ALLOWED; i++) {
+                                        int returnCode = kmAttachmentsController.processFile(clipData.getItemAt(i).getUri(), alCustomizationSettings, prePostUIMethods);
+                                        doReturnCodeActions(returnCode);
+                                    }
+                                }
+                            });
+                            dialogFragmentSelectLimitCrossed.show(getFragmentManager(), "FragmentSelectLimitCrossed");
```
reytum: Move this string to strings file as this is part of localisation

shubhamtewari: This is just the fragment tag, won't be visible to the user. Do we still need to localize it?

PR #78, opened 2019-10-29
add launchAndCreateIfEmpty method in conversation builder

reytum:
Added a new method in KmConversationBuilder that will launch:
1) Launch the conversation list if multiple conversations are there
2) Open the conversation thread in case of a single conversation
3) Create and launch the conversation in case if there are no conversations for the user

All the properties are taken from KmConversationBuilder object.

It works as below:
```java
new KmConversationBuilder(MainActivity.this)
                        .setAppId(APP_ID)
                        .launchAndCreateIfEmpty(new KmCallback() {
                            @Override
                                Utils.printLog(MainActivity.this, "LaunchTest", "Launch and create if empty success : " + message);
                            }

                            @Override
                                Utils.printLog(MainActivity.this, "LaunchTest", "Launch and create if empty error : " + error);
                            }
                        });
```

This is tested on local device with visitor login.

### DIFF
--- a/kommunicate/src/main/java/io/kommunicate/KmConversationHelper.java
+++ b/kommunicate/src/main/java/io/kommunicate/KmConversationHelper.java
@@ -419,18 +419,13 @@ public class KmConversationHelper {
             @Override
             public void onResult(List<Message> messageList, ApplozicException e) {
                 if (e == null) {
-                    if (messageList != null) {
-                        if (messageList.size() == 0) {
-                            conversationBuilder.setSkipConversationList(false);
-                            conversationBuilder.launchConversation(callback);
-                        } else if (messageList.size() == 1) {
-                            openParticularConversation(conversationBuilder.getContext(), false, messageList.get(0).getGroupId(), callback);
-                        } else if (messageList.size() > 1) {
-                            Kommunicate.openConversation(conversationBuilder.getContext(), callback);
-                        }
-                    } else {
+                    if (messageList.isEmpty()) {
                         conversationBuilder.setSkipConversationList(false);
                         conversationBuilder.launchConversation(callback);
+                    } else if (messageList.size() == 1) {
+                        openParticularConversation(conversationBuilder.getContext(), false, messageList.get(0).getGroupId(), callback);
+                    } else {
+                        Kommunicate.openConversation(conversationBuilder.getContext(), callback);
                     }
                 }
             }
--- a/kommunicate/src/main/java/io/kommunicate/KmConversationHelper.java
+++ b/kommunicate/src/main/java/io/kommunicate/KmConversationHelper.java
@@ -55,7 +55,7 @@ public class KmConversationHelper {
 
     public static void openConversation(final Context context, final boolean skipConversationList, final Integer conversationId, final KmCallback callback) throws KmException {
         if (!(context instanceof Activity)) {
-            throw new KmException("This method needs Activity context");
+            throw new KmException(Utils.getString(context, R.string.km_method_needs_activity_context));
         }
 
         if (conversationId == null) {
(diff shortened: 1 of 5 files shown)


kommunicate/src/main/java/io/kommunicate/KmConversationHelper.java
```diff
@@ -385,70 +388,53 @@ public void onReceive(KMUser user, ResultReceiver resultReceiver) {
...
+            @Override
+            public void onResult(List<Message> messageList, ApplozicException e) {
+                if (e == null) {
+                    if (messageList != null) {
+                        if (messageList.size() == 0) {
+                            conversationBuilder.setSkipConversationList(false);
+                            conversationBuilder.launchConversation(callback);
+                        } else if (messageList.size() == 1) {
+                            openParticularConversation(conversationBuilder.getContext(), false, messageList.get(0).getGroupId(), callback);
+                        } else if (messageList.size() > 1) {
```
devashishmamgain: condition inside if is not required here

kommunicate/src/main/res/values/strings.xml
```diff
@@ -3,4 +3,13 @@
     <string name="unable_to_create_conversation">Unable to create conversation</string>
     <string name="km_prechat_processing_wait_info">Please wait, launching conversation..</string>
     <string name="applozic_userId_error_info_in_logs">UserId can not be null or empty and channel key should not be null</string>
+
+    <!--Kommunicate conversation errors-->
+    <string name="km_method_needs_activity_context">This method needs Activity context</string>
```
devashishmamgain: is this for log or to show to end user?

reytum: This is for log

kommunicate/src/main/java/io/kommunicate/KmConversationHelper.java
```diff
@@ -385,70 +388,53 @@ public void onReceive(KMUser user, ResultReceiver resultReceiver) {
...
-                        public void onReceive(KMUser user, ResultReceiver resultReceiver) {
-                            Kommunicate.login(conversationBuilder.getContext(), user, getLoginHandler(conversationBuilder, getStartConversationHandler(conversationBuilder.isSkipConversationList(), false, resultReceiver, callback), callback));
+            return;
+        }
+
+        ApplozicConversation.getLatestMessageList(conversationBuilder.getContext(), false, new MessageListHandler() {
+            @Override
+            public void onResult(List<Message> messageList, ApplozicException e) {
+                if (e == null) {
+                    if (messageList != null) {
```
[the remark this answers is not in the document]
reytum: messageList will not be null.
But we need to check the size of the list so that we can do those actions based on the size of the list.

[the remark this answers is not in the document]
devashishmamgain: Then we can check for isempty and size 1

kommunicate/src/main/java/io/kommunicate/KmConversationHelper.java
reytum: Done

kommunicate/src/main/java/io/kommunicate/KmConversationHelper.java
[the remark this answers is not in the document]
reytum: Done

kommunicate/src/main/java/io/kommunicate/KmConversationHelper.java
```diff
@@ -385,70 +388,48 @@ public void onReceive(KMUser user, ResultReceiver resultReceiver) {
...
+        }
 
-                if (conversationBuilder.getKmUser() != null) {
-                    kmUser = conversationBuilder.getKmUser();
-                } else {
-                    kmUser = Kommunicate.getVisitor();
+        ApplozicConversation.getLatestMessageList(conversationBuilder.getContext(), false, new MessageListHandler() {
+            @Override
+            public void onResult(List<Message> messageList, ApplozicException e) {
+                if (e == null) {
```
devashishmamgain: we can print log statement in case of exception, currently if any error comes, we wouldn't be able to figure out the issue.

reytum: Ok, will add logs as well, but the errors are also sent in failure callback.

PR #89, opened 2019-12-09
Rich message rendering fix on sender side

reytum:
1) Fixed The rich message UI issue where the messages were not being rendered on the sender side.
2) Created 2 seperate layouts for sender and receiver side of the rich messages as both of them use different margins.
3) Did some code optimisation in KmCustomFlowLayout . This is the layout which renders the quick replies layout. 
4) Disabled the global storage permission by default.
5) Added deep linking feature for weblinks

Tested on local device


kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/AlCustomizationSettings.java
```diff
@@ -106,7 +106,7 @@
     private String awayMessageTextColor = "#A9A4A4";
     private boolean isAgentApp = false;
     private boolean hideGroupSubtitle = false;
-    private boolean disableGlobalStoragePermission = false;
+    private boolean disableGlobalStoragePermission = true;
```
devashishmamgain: can we change it to the other globalGlobalStoragePermission true or false.
its easy to understand something as:
enable: true vs disable: false  - although both means the same thing

kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/conversation/richmessaging/views/KmFlowLayout.java
```diff
@@ -0,0 +1,62 @@
+package com.applozic.mobicomkit.uiwidgets.conversation.richmessaging.views;
+
+import android.content.Context;
+import android.util.AttributeSet;
+import android.view.ViewGroup;
+
+public abstract class KmFlowLayout extends ViewGroup {
+
+    protected int line_height_space;
```
devashishmamgain: some places we have _ and other places camelCase, is there any specific reason for using _ here?

kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/conversation/richmessaging/views/KmReceiverFlowLayout.java
```diff
@@ -0,0 +1,119 @@
...
+        if (isRtlDirection) {
+            xpos = width - getPaddingStart();
+        }
+        for (int i = 0; i < getChildCount(); i++) {
+            final View child = getChildAt(i);
+            if (child.getVisibility() != GONE) {
+                final int childw = child.getMeasuredWidth();
+                final int childh = child.getMeasuredHeight();
+                final KmReceiverFlowLayout.LayoutParams lp = (KmReceiverFlowLayout.LayoutParams) child.getLayoutParams();
+                if (isRtlDirection) {
```
devashishmamgain: this looks similar to the code added in processOnMeasure, can we move the common part to a method which takes the values and return the final xpos, pos

kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/conversation/richmessaging/views/KmSenderFlowLayout.java
```diff
@@ -0,0 +1,117 @@
...
+        for (int i = 0; i < getChildCount(); i++) {
+            final View child = getChildAt(i);
+            if (child.getVisibility() != GONE) {
+                final LayoutParams lp = (LayoutParams) child.getLayoutParams();
+                child.measure(MeasureSpec.makeMeasureSpec(width, View.MeasureSpec.AT_MOST), childHeightMeasureSpec);
+                final int childw = child.getMeasuredWidth();
+                line_height_space = Math.max(line_height_space, child.getMeasuredHeight() + lp.vertical_spacing);
+
+                if (isNormalDirection) {
+                    if (childw > xpos) {
```
devashishmamgain: looks similar code as above

kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/conversation/richmessaging/views/KmFlowLayout.java
reytum: Will change it to camelCase

kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/AlCustomizationSettings.java
reytum: This setting will not be needed now, we will remove this from docs. The storage permission is not needed globally when we have a better implementation of checking them on demand. Thats why we are setting it as true internally so that users don't have to change it.

PR #92, opened 2019-12-31
Add option to set dynamic parent activity for ConversationActivity

reytum:
1) Added option to set parent activity dynamically
2) Fixed the submit/quick reply button click in case type is not specified
3) Set the default group name to "Support" from "Kommunicate Support"
4) Fixed the sender side text color in document view
5) Fixed the FAQ button click

### DIFF
--- a/kommunicate/src/main/java/io/kommunicate/Kommunicate.java
+++ b/kommunicate/src/main/java/io/kommunicate/Kommunicate.java
@@ -306,7 +306,7 @@ public class Kommunicate {
     private static void createConversation(KmChatBuilder chatBuilder, KMStartChatHandler handler) throws KmException {
         List<KMGroupInfo.GroupUser> users = new ArrayList<>();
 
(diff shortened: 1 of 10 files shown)


kommunicate/src/main/java/io/kommunicate/Kommunicate.java
```diff
@@ -306,7 +306,7 @@ public void onFailure(Object error) {
     private static void createConversation(KmChatBuilder chatBuilder, KMStartChatHandler handler) throws KmException {
         List<KMGroupInfo.GroupUser> users = new ArrayList<>();
 
-        KMGroupInfo channelInfo = new KMGroupInfo(TextUtils.isEmpty(chatBuilder.getChatName()) ? "Kommunicate Support" : chatBuilder.getChatName(), new ArrayList<String>());
+        KMGroupInfo channelInfo = new KMGroupInfo(TextUtils.isEmpty(chatBuilder.getChatName()) ? "Support" : chatBuilder.getChatName(), new ArrayList<String>());
```
devashishmamgain: is this the default name? we can make it same as web, i guess it is "Conversations" on web. @Rudranand pls confirm

reytum: done

PR #94, opened 2020-01-02
Add is logged in check in login method

reytum:
1) Added the islogged in check in the login method internally to avoid multiple login calls.
2) User will still get the callback in onSuccess of login with messsage saying USER ALREADY LOGGED IN and some other user details that are saved on device.

Tested on local device


kommunicate/src/main/java/io/kommunicate/Kommunicate.java
```diff
@@ -71,7 +73,22 @@ public static void init(Context context, String applicationKey) {
     }
 
     public static void login(Context context, KMUser kmUser, KMLoginHandler handler) {
-        login(context, kmUser, handler, null);
+        if (isLoggedIn(context)) {
```
devashishmamgain: what will happen if someone will pass a different userId?

reytum: Did the changes for that.

PR #101, opened 2020-01-22
Show toast for App id change 

reytum:
1)Show toast and log for app ID change in Kommunicate.init() function
2) Show error and toast for empty app ID

Tested on local device.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -6,5 +6,5 @@
     <string name="some_error_occured">Some error occurred</string>
     <string name="login_as_visitor_info">When logging in as visitor, you don\'t need to fill the email and password fields.
         Clicking the \'Login as visitor\' button will log you in with a random userId</string>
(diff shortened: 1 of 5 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -6,5 +6,5 @@
     <string name="some_error_occured">Some error occurred</string>
     <string name="login_as_visitor_info">When logging in as visitor, you don\'t need to fill the email and password fields.
         Clicking the \'Login as visitor\' button will log you in with a random userId</string>
-    <string name="inavild_app_id_error">Looks like you have not added your application Id in your app level build.gradle file. Please provide your application ID in app level build.gradle file\'s defaultConfig <a href='https://github.com/Kommunicate-io/Kommunicate-Android-Chat-SDK'>Refer here</a>. Replace Your-APP_ID with the application ID obtained from Kommunicate dashboard.</string>
+    <string name="inavild_app_id_error">Looks like you have not added your application Id in your app level build.gradle file. Please provide your application ID in app level build.gradle file\'s defaultConfig <a href='https://github.com/Kommunicate-io/Kommunicate-Android-Chat-SDK'>Refer here</a>. Replace &lt;Your-APP-ID&gt; with the application ID obtained from Kommunicate dashboard.</string>
```
devashishmamgain: inavild -> invalid

devashishmamgain: is Your-APP-ID mentioned somewhere in repo or code?

reytum: <Your-APP-ID> is the place holder for app ID in sample app's build.gradle file. It will only be used when someone is using our sample app for testing.

PR #108, opened 2020-02-07
[CM-97] Add speech to text and text to speech support

reytum:
* To enable the settings:
    
```
    KmPrefSettings.getInstance(SplashScreenActivity.this)
                        .enableSpeechToText(true)
                        .enableTextToSpeech(true)
                        .setSendMessageOnSpeechEnd(true)
                        .setSpeechToTextLanguage("zh-TW")
                        .setTextToSpeechLanguage("zh-TW");
```
* Text to speech:  Works only when the conversation is open and a new message is received within the conversation. 

* Speech to text: The mic button used to record and send audio will function as STT starter. Just click the button once and the device will start listening. To cancel an ongoing listening, click on the button again. The audio button will scale animate depending on the input voice amplitude. The partial results will be set on the message edit text but the send message button will only be enabled once the complete speech is received. When `isSendOnSpeechEnd = true` , the message will be sent automatically once the text is received.

* To Integrate your own STT or TTS service, replace the KmSpeechToText and KmTextToSpeech implementation with your service.


kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/kommunicate/KmPrefSettings.java
```diff
@@ -0,0 +1,124 @@
...
+    private static final String ENABLE_FAQ_OPTION = "ENABLE_FAQ_OPTION";
+    private static final String ENABLE_SPEECH_TO_TEXT = "ENABLE_SPEECH_TO_TEXT";
+    private static final String ENABLE_TEXT_TO_SPEECH = "ENABLE_TEXT_TO_SPEECH";
+    private static final String SEND_MESSAGE_ON_SPEECH_END = "SEND_MESSAGE_ON_SPEECH_END";
+    private static final String SPEECH_TO_TEXT_LANGUAGE = "SPEECH_TO_TEXT_LANGUAGE";
+    private static final String TEXT_TO_SPEECH_LANGUAGE = "TEXT_TO_SPEECH_LANGUAGE";
+
+    private KmPrefSettings(Context context) {
+        if (!TextUtils.isEmpty(MobiComKitClientService.getApplicationKey(context))) {
+            sharedPreferences = context.getSharedPreferences(MobiComKitClientService.getApplicationKey(context), Context.MODE_PRIVATE);
```
Sunilkumarr: `sharedPreferences` we can check if we can create with some constant  name  and clear in logout

reytum: Its already live with appId name. If we change it, it might break the existing application.

Sunilkumarr: okay

PR #115, opened 2020-02-26
[CM-173] Option to set custom fields in Prechat Screen

reytum:
Users can now set custom fields in the prechat screen. The field types currently supported are : email, text, number and password. Use the code as below:

```java
  List<KmPrechatInputModel> inputModelList = new ArrayList<>();  //create list of input fields
  
//add fields
 inputModelList.add(new KmPrechatInputModel()
                        .setField("Email") //This will be the label for the field
                        .setType(KmPrechatInputModel.KmInputType.EMAIL) //type
                        .setValidationError("Kripaya is bharein") //Error displayed if text entered is invalidation
                        .setValidationRegex("Some email regex") //Regex to validate text
                        .setCompositeRequiredField("Mobile number");  //used to set an option of either of the 2 fileds required. This means that the user will have to enter either email or phone number to proceed

inputModelList.add(new KmPrechatInputModel()
                        .setField("Mobile number")
                        .setType(KmPrechatInputModel.KmInputType.NUMBER)
                        .setValidationError("Kripaya is bharein")
                        .setValidationRegex("Some number regex");

inputModelList.add(new KmPrechatInputModel()
                        .setField("Password")
                        .setType(KmPrechatInputModel.KmInputType.PASSWORD)
                        .setValidationError("Kripaya is bharein")
                        .setValidationRegex("Some regex")
                        .setRequired(true));  //This makes this field mandatory

Kommunicate.launchPrechatWithResult(SplashScreenActivity.this, inputModelList, new KmPrechatCallback<Map<String, String>>() {
                    @Override
                    public void onReceive(Map<String, String> data, Context context, ResultReceiver finishActivityReceiver) {
                       //data map will receive the "filedName: entered Text" map. You can get the texts and create a KMUser object here. Then call KmConversationBuilder with that user object.
                        Utils.printLog(context, "TestPre", GsonUtils.getJsonFromObject(data, Map.class));
                    }

                    @Override
                    public void onError(String error) {
                        Utils.printLog(SplashScreenActivity.this, "TestPre", "Error : " + error);
                    }
                });
```


kommunicateui/src/main/res/values/mobicom_strings.xml
```diff
@@ -239,7 +239,7 @@
     <string name="nameEt">Name</string>
     <string name="phoneNumberEt">Phone number</string>
     <string name="prechat_screen_text">We just need a few details to help you get started</string>
-    <string name="prechat_screen_toast_error_message">Either contact number or emailId is required.</string>
+    <string name="prechat_screen_toast_error_message">Either %1$s or %2$s is required.</string>
```
devashishmamgain: remove .

reytum: Done

PR #121, opened 2020-03-20
[CM-13] Option to enable themes from dashboard

reytum:
1) Now the primary color can be set from the dashboard, It will be updated across all toolbars, sent message background, status icons, some rich message texts and backgrounds etc
2) No option to set secondary color from dashboard. But incase the secondaryColor is received, it will be set in the status bar.
3) The priority goes as below:
    applozic-settings.json -> chatWidget theme from dashboard -> default colors file
4) Currently, the update will only happen incase of logout -> login, but the task is designed to easily work with real-time as well in future.


kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/kommunicate/utils/KmThemeHelper.java
```diff
@@ -0,0 +1,110 @@
...
+
+    public int getPrimaryColor() {
+        if (primaryColor == -1) {
+            String colorStr = appSettingPreferences.getPrimaryColor();
+            primaryColor = !TextUtils.isEmpty(colorStr) ? Color.parseColor(colorStr) : context.getResources().getColor(R.color.applozic_theme_color_primary);
+        }
+        return primaryColor;
+    }
+
+    public int getSentMessageBackgroundColor() {
```
devashishmamgain: @reytum this and below 2 methods code looks similar, is it possible to find a way to create a single method that can work for all the 3?

kommunicate/src/main/java/io/kommunicate/models/KmAppSettingModel.java
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    public class KmChatWidget extends JsonMarker {
+        private boolean popup;
+        private Short iconIndex;
+        private String primaryColor;
+        private String secondaryColor;
+        private boolean showPoweredBy;
+        private long sessionTimeout;
+        private long botMessageDelayInterval;
```
devashishmamgain: it is possible that customer havn't configured it so this can be null, same with sessionTimeout.
safe side, it will be better to use wrapper class.
Also, these 2 value would never be that big to come under long.

kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/AlCustomizationSettings.java
```diff
@@ -12,9 +12,9 @@
 public class AlCustomizationSettings extends JsonMarker {
 
     private String customMessageBackgroundColor = "#e6e5ec";
-    private String sentMessageBackgroundColor = "#5c5aa7";
+    private String sentMessageBackgroundColor = "";
```
devashishmamgain: worst case if server API fails, anything will break on design?

kommunicate/src/main/java/io/kommunicate/async/KmAppSettingTask.java
```diff
@@ -0,0 +1,53 @@
...
+        this.context = new WeakReference<>(context);
+        this.appId = appId;
+        this.callback = callback;
+    }
+
+    @Override
+    protected KmAppSettingModel doInBackground(Void... voids) {
+        try {
+            KmAppSettingPreferences.getInstance().clearInstance();
+            return KmAppSettingPreferences.fetchAppSetting(context.get(), appId);
```
devashishmamgain: @reytum would it be better if we fetch the settings first and check that the call was successful and then call .clearInstance()? in case if app settings fail?

kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/kommunicate/utils/KmThemeHelper.java
reytum: They look same because we are setting only one color for all. If customer wants to set different color then different methods are required. Seperate methods are there because these colors are different in applozic-settings.json file. This is for future, if we provide option to set different colors.

kommunicate/src/main/java/io/kommunicate/models/KmAppSettingModel.java
reytum: I guess sessionTimeout and botMessageDelay are not required on mobile side and are neither used currenlty. We can remove these properties.

kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/AlCustomizationSettings.java
reytum: Nothing will break as the fallback color is from the colors.xml file which will always be there.
The sequence is:
```
   if(color is set in applozic-settings) {
      use this color
   } else if(color received from server){
      use server color
} else {
    use the default color from colors.xml file
}
```

kommunicate/src/main/java/io/kommunicate/async/KmAppSettingTask.java
reytum: We could do that.

kommunicate/src/main/java/io/kommunicate/models/KmAppSettingModel.java
devashishmamgain: they will be used in later versions, we can keep them

reytum: Yes, if those woudnt be timestamp then we can change them to int.

devashishmamgain: it will be integers, 
bot time delay is in set in seconds on dashboard
and no one will set it to be more than 15-20 seconds. Standard value for this is 1 second typing time.
trying setting the value here: https://dashboard.kommunicate.io/settings/chat-widget-customization#bot-typing-indicator
and
https://dashboard.kommunicate.io/settings/chat-widget-customization#chat-widget-security
to see the response value coming.
check in both cases, one is a new signup without modifying any value. another after setting some value.

reytum: Done

kommunicate/src/main/java/io/kommunicate/async/KmAppSettingTask.java
reytum: Done

devashishmamgain: is it committed? code looks same as before

PR #123, opened 2020-03-30
Added way to sending usersIds from builder.

Sunilkumarr:
* Added a way for adding more users in the group using the builder so that they can add users other than the agent Ids or bot ids.

* Tested with and without adding members in the group works fine.

Testing code: 
 ```
    List<String> agentIds = new ArrayList<>();
                    agentIds.add("<AGENT_ID>"); // Add AGENT_ID(s) to this list.
                    List<String> botIds = new ArrayList<>();
                    botIds.add("BOT_ID");// Add BOT_ID(s) to this list.

                    List<String> userIds = new ArrayList<>();
                    userIds.add("userId1");// Add USER id to this list.
                    userIds.add("userId2");// Add USER id  to this list.

                    new KmConversationBuilder(MainActivity.this)
                            .setAgentIds(agentIds)
                            .setBotIds(botIds)
                            .setUserIds(userIds)
                            .createConversation(new KmCallback() {
                                @Override

                                }

                                @Override

                                }
                            });
```

1. Test without using  ```.setUserIds(userIds)``` if this is not added it should work as it was working before 
2. Test with using  ```.setUserIds(userIds)``` if this is added then users need be in that support group the user can chat in that support group as a group.

### DIFF
--- a/kommunicate/src/main/java/io/kommunicate/KmConversationBuilder.java
+++ b/kommunicate/src/main/java/io/kommunicate/KmConversationBuilder.java
@@ -20,6 +20,7 @@ public class KmConversationBuilder extends JsonMarker {
     private String appId;
     private List<String> agentIds;
     private List<String> botIds;
+    private List<String> userIds;
     private boolean skipConversationList = true;
     private String conversationId;
     private String fcmDeviceToken;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: kommunicate/src/main/java/io/kommunicate/KmConversationBuilder.java [hunk 1]
```diff
@@ -199,6 +200,15 @@ public KmConversationBuilder setConversationTitle(String conversationTitle) {
         return this;
     }
 
+    public List<String> getUserIds() {
+        return userIds;
+    }
+
+    public KmConversationBuilder setUserIds(List<String> userIds) {
```
devashishmamgain: @Sunilkumarr pls check if builder pattern uses method names as setX or withX?

Sunilkumarr: Mostly I see with setX in android this are the links for  example 
1.  https://dzone.com/articles/the-builder-pattern-for-class-with-many-constructo
2. https://www.vogella.com/tutorials/DesignPatternBuilder/article.html

reytum:
Change the base branch to v1.9.7

PR #130, opened 2020-05-28
Add support for language change bot

reytum:
Added the option to change the user language from the click of language selection quick replies.


kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/conversation/richmessaging/RichMessageActionProcessor.java
```diff
@@ -161,7 +170,9 @@ public void handleQuickReplies(Object object, Map<String, Object> replyMetadata)
             }
         } else if (object instanceof ALRichMessageModel.AlAction) {
             ALRichMessageModel.AlAction action = (ALRichMessageModel.AlAction) object;
+            updateLanguage(action.getUpdateLanguage());
```
Sunilkumarr, line 173: We can move this `updateLanguage`  method to the bottom of this `else if` and store the language in the string  and use it

PR #131, opened 2020-06-08
CM-270: add away status for agent

shubhamtewari:


shubhamtewari:
CM-270: add away status for agent.

PR #132, opened 2020-06-16
CM-291: add message text character limit for dialog flow bot

shubhamtewari:


kommunicate/src/main/java/io/kommunicate/async/KmGetBotTypeTask.java
```diff
@@ -0,0 +1,75 @@
...
+                        callback.onSuccess(botType);
+                    }
+                } catch (Exception exception) {
+                    callback.onFailure(exception.getMessage());
+                }
+            }
+        }
+    }
+
+    public static class BotDetailsResponseData {
```
devashishmamgain, line 62: is class required for this? or enum can be used

shubhamtewari: yes. it's being used as a POJO class to deserialize the response json.

PR #143, opened 2020-07-13
CM-310: unique conversation setting from server

shubhamtewari:


kommunicate/src/main/java/io/kommunicate/KmConversationHelper.java
```diff
@@ -349,6 +349,21 @@ public void onFailure(RegistrationResponse registrationResponse, Exception excep
         };
     }
 
+    //meant to be used from the conversation screen start new conversation button
+    public static void launchConversationIfLoggedIn(Context context, KmCallback callback) {
+        if (Kommunicate.isLoggedIn(context)) {
+            KmConversationBuilder conversationBuilder = new KmConversationBuilder(context).setAgentIds(null).setBotIds(null);
```
devashishmamgain, line 355: whats the need to set null?

kommunicate/src/main/java/io/kommunicate/models/KmAppSettingModel.java
```diff
@@ -116,5 +117,9 @@ public int getBotMessageDelayInterval() {
         public void setBotMessageDelayInterval(int botMessageDelayInterval) {
             this.botMessageDelayInterval = botMessageDelayInterval;
         }
+
+        public boolean isSingleThreaded() {
```
devashishmamgain, line 121: wrong method. if value is x, then method should isX
in this case, value is isX and method is also isX

shubhamtewari: fixed

kommunicate/src/main/java/io/kommunicate/KmConversationHelper.java
shubhamtewari: removed

kommunicate/src/main/java/io/kommunicate/KmConversationHelper.java
```diff
@@ -667,43 +682,47 @@ public void onFailure(ChannelFeedApiResponse channelFeedApiResponse, Context con
...
+                KmAppSettingModel.KmResponse kmAppSettings = (KmAppSettingModel.KmResponse) message;
+                if (kmAppSettings != null) {
+                    List<String> agents = new ArrayList<>();
+                    agents.add(kmAppSettings.getAgentId());
+                    conversationBuilder.setAgentIds(agents);
+                    if (useSingleThreadedSettingFromServer) {
+                        conversationBuilder.setSingleConversation(kmAppSettings.getChatWidget().isSingleThreaded());
+                    }
+                    try {
+                        final String clientChannelKey = !TextUtils.isEmpty(conversationBuilder.getClientConversationId()) ? conversationBuilder.getClientConversationId() : (conversationBuilder.isSingleConversation() ? getClientGroupId(conversationBuilder.getUserIds(), agents, conversationBuilder.getBotIds(), conversationBuilder.getContext()) : null);
```
reytum, line 698: The priority is given to clientConversationId over single conversation if it is passed in the builder. Put this check above. SO if the builder has client conversation Id , there is no need to make the app setting call. Its an unnecessary call. Also the app setting call is already done once to get the agents, We can merge these two calls and also process the theme on success. This will not only reduce the duplicate call but also change the theme without need to logout /login.

kommunicate/src/main/java/io/kommunicate/KmConversationHelper.java
```diff
@@ -667,43 +682,47 @@ public void onFailure(ChannelFeedApiResponse channelFeedApiResponse, Context con
...
+        };
+    }
 
-            new KmGetAgentListTask(conversationBuilder.getContext(), MobiComKitClientService.getApplicationKey(conversationBuilder.getContext()), callback).execute();
+    private static void startConversation(boolean useSingleThreadedSettingFromServer, final KmConversationBuilder conversationBuilder, final KmStartConversationHandler handler) throws KmException {
+        if (conversationBuilder == null) {
+            throw new KmException(Utils.getString(conversationBuilder.getContext(), R.string.km_conversation_builder_cannot_be_null));
+        }
+        if (conversationBuilder.getAgentIds() == null || conversationBuilder.getAgentIds().isEmpty()) {
+            new KmGetAgentListTask(conversationBuilder.getContext(), MobiComKitClientService.getApplicationKey(conversationBuilder.getContext()), getCallbackWithAppSettingsToCreateConversation(useSingleThreadedSettingFromServer, conversationBuilder, handler)).execute();
```
shubhamtewari, line 725: This is the call. @reytum

PR #152, opened 2020-08-12
Add support for date and time picker in form template

reytum:
* Added support for three types: date, time and datetime-local
* Added support for 24hrs and 12hrs in time

Sample JSON:

```
{
"payload": [
      {
        "type": "date",     // for date
        "data": {
          "label": "Date"
        }
      }, 
      {
        "type": "time",  //for time
        "data": {
          "label": "Time",
           "amPm": true.      //optional field to set 24 hr and 12 hr time. Set false for 24 hrs
        }
      },
      {
        "type": "datetime-local",  //for time
        "data": {
          "label": "Date & Time",
           "amPm": true.      //optional field to set 24 hr and 12 hr time. Set false for 24 hrs
        }
    ]
}
```

### DIFF
--- a/kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/conversation/richmessaging/adapters/KmFormItemAdapter.java
+++ b/kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/conversation/richmessaging/adapters/KmFormItemAdapter.java
@@ -1,6 +1,7 @@
 package com.applozic.mobicomkit.uiwidgets.conversation.richmessaging.adapters;
 
 import android.app.DatePickerDialog;
+import android.app.TimePickerDialog;
 import android.content.Context;
 import android.text.Editable;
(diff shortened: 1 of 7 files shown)


kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/conversation/richmessaging/adapters/KmFormItemAdapter.java
```diff
@@ -215,15 +234,99 @@ public void onCheckedChanged(CompoundButton buttonView, boolean isChecked) {
                         } else {
                             formItemViewHolder.flowLayout.setVisibility(View.GONE);
                         }
+                    } else if (KmFormPayloadModel.Type.DATE.getValue().equals(payloadModel.getType())
```
Sunilkumarr, line 237: inside payload model, we can move this type checking

PR #158, opened 2020-10-07
CM-477: change toolbar text color from settings file

shubhamtewari:
Change:
To change the conversation screen toolbar title and subtitle color now you can add "toolbarTitleColor" and "toolbarSubtitleColor" fields to the applozic-settings.json file. Any error in entry will result in the default being white.
**_Tested locally_**, by going through a simple flow to customize the color, as it would likely be followed by a customer.

Note: Working on moving the toolbar logic to new class. This doesn't concern the issue CM-477 and hasn't been pushed now. It's WIP.

### DIFF
--- a/kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/conversation/activity/ConversationActivity.java
+++ b/kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/conversation/activity/ConversationActivity.java
@@ -365,6 +365,7 @@ public class ConversationActivity extends AppCompatActivity implements MessageCo
         KmUtils.setStatusBarColor(this, KmThemeHelper.getInstance(this, alCustomizationSettings).getSecondaryColor());
         customToolbarLayout = myToolbar.findViewById(R.id.custom_toolbar_root_layout);
         setSupportActionBar(myToolbar);
+        setToolbarTitleSubtitleColorFromSettings();
         baseContactService = new AppContactService(this);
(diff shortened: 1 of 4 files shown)


kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/conversation/activity/ConversationActivity.java
```diff
@@ -681,6 +682,29 @@ void beginCrop(Uri imageUri) {
...
+            return Color.WHITE;
+        }
+    }
+
+    public void setToolbarTitleSubtitleColorFromSettings() {
+        if(customToolbarLayout == null) {
+            return;
+        }
+
+        int titleColor = parseColorWithDefaultWhite(alCustomizationSettings.getToolbarTitleColor());
```
reytum, line 698: Use KmThemeHelper here. Access the colors from Theme helper.

Also fix the bug in the same file here: https://github.com/Kommunicate-io/Kommunicate-Android-Chat-SDK/blob/9851b35405f1fd3ee7b2d15db2f361d676621b01/kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/kommunicate/utils/KmThemeHelper.java#L62. change sentMessageBackgroundCOlor to sentMessageBorderColor

shubhamtewari: done

reytum (approved):

PR #169, opened 2020-11-26
rename all Al/Applozic place holder to Km/Kommunicate

reytum:


shubham-applozic:
AlCustomizationSettings. Should this be renamed too?

shubham-applozic (approved):

reytum:
> AlCustomizationSettings. Should that be renamed too?

For the time being we will keep it. In next sprint we will add kommunicate-settings.jsonfile support as well in the AL SDK. Then we can rename it.

PR #184, opened 2021-03-04
update away message on assignee change

reytum:
* Fetch away message whenever the assignee changes

### DIFF
--- a/kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/conversation/fragment/MobiComConversationFragment.java
+++ b/kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/conversation/fragment/MobiComConversationFragment.java
@@ -1075,11 +1075,27 @@ public abstract class MobiComConversationFragment extends Fragment implements Vi
             return;
         }
 
-        channel = ChannelService.getInstance(getActivity()).getChannelByChannelKey(channel.getKey());
+        String existingAssignee = channel.getConversationAssignee();
 
+        channel = ChannelService.getInstance(getActivity()).getChannelByChannelKey(channel.getKey());
         //conversation is open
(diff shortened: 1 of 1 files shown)


kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/conversation/fragment/MobiComConversationFragment.java
```diff
@@ -1075,11 +1075,27 @@ public void onChannelUpdated() {
...
 
-        channel = ChannelService.getInstance(getActivity()).getChannelByChannelKey(channel.getKey());
+        String existingAssignee = channel.getConversationAssignee();
 
+        channel = ChannelService.getInstance(getActivity()).getChannelByChannelKey(channel.getKey());
         //conversation is open
         //if the conversation is opened from the dashboard while the feedback input fragment is open, the feedback fragment will be closed
         setFeedbackDisplay(channel != null && channel.getKmStatus() == Channel.CLOSED_CONVERSATIONS && !KmUtils.isAgent(getContext()));
+
+        if (existingAssignee != null && !existingAssignee.equals(channel.getConversationAssignee())) {
```
devashishmamgain, line 1085: use !channel.getConversationAssignee().equals(...) then null check will not be required

devashishmamgain (changes requested):

kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/conversation/fragment/MobiComConversationFragment.java
reytum: channel.getConversationAssignee() might also be null, that's why placed it inside so that we can avoid that null check.

devashishmamgain (approved):

PR #185, opened 2021-03-04
Add setting to disable message sending option for bot

reytum:


app/src/main/assets/applozic-settings.json
```diff
@@ -85,6 +85,7 @@
   "toolbarColor": "",
   "statusBarColor": "",
   "richMessageThemeColor": "",
+  "hideKeyboardForBot": false,
```
devashishmamgain, line 88: are we hiding only keyboard while continuing to show the textfield? or we are hiding the whole?

devashishmamgain (changes requested):

app/src/main/assets/applozic-settings.json
reytum: Hiding the whole. The keyboard would hide automatically once the text field is hidden

reytum: Screenshot_20210304-212004

devashishmamgain:
Any other generic name for setting which can be used in web also for this
feature?

reytum:
How about disableCustomMessageForBots or disableSendMessageForBots ?

devashishmamgain:
restrictMessageTyingWithBots ?

reytum:
Looks good.

devashishmamgain (approved):

devashishmamgain (approved):

PR #191, opened 2021-04-01
add functionality to update conversation metadata and assignee

reytum:
* Added functionality to update conversation metadata via conversation builder
* Added functionality to update conversationAssignee if a different assignee is passed in the conversation builder
* Moved the updates into parallel tasks so it doesn't affect the response time of conversation builder
* Use the below method to update conversation metadata:

```java
  Map<String, String> map = new HashMap<>();
            map.put("TicketId", "123456");
            map.put("Status", "ActiveNo");

            new KmConversationBuilder(context)
                    //other properties might go here
                    .setConversationAssignee(newAssignee)
                    .setConversationMetadata(map)
                    .launchConversation(new KmCallback() {
                        @Override
            
                        }

                        @Override
                            
                        }
                    });
```

Below test cases are satisfied:
1) If this is a new conversation, the conversation will be created and the above metadata will be set.
2) If this is an existing conversation, this metadata will be set, if conversation does not have metadata this metadata
3) If the values change or new metadata is passed, the metadata will be updated accordingly.
4) If same metadata is passed, no update occurs.

### DIFF
--- a/kommunicate/src/main/java/com/applozic/mobicomkit/api/HttpRequestUtils.java
+++ b/kommunicate/src/main/java/com/applozic/mobicomkit/api/HttpRequestUtils.java
@@ -36,220 +36,136 @@ public class HttpRequestUtils {
     public static String DEVICE_KEY_HEADER = "Device-Key";
     private static final String APZ_PRODUCT_APP_HEADER = "Apz-Product-App";
     public static boolean isRefreshTokenInProgress = false;
+    private String encryptionKey;
     private Context context;
 
 
     public HttpRequestUtils(Context context) {
         this.context = ApplozicService.getContext(context);
+        this.encryptionKey = MobiComUserPreference.getInstance(context).getEncryptionKey();
+    }
+
+    public String postData(String urlString, String data) throws Exception {
+        return postData(urlString, data, null, false, false, null, null);
+    }
+
+    public String postData(String urlString, String data, String userId) throws Exception {
+        return postData(urlString, data, userId, false, false, null, null);
     }
 
     public String postData(String urlString, String contentType, String accept, String data) throws Exception {
-        return postData(urlString, contentType, accept, data, null);
-    }
-
-    public String postDataForAuthToken(String urlString, String contentType, String accept, String data, String userId) throws Exception {
-        Utils.printLog(context, TAG, "Calling url: " + urlString);
-        HttpURLConnection connection;
-        URL url;
-        try {
-            if (!TextUtils.isEmpty(MobiComUserPreference.getInstance(context).getEncryptionKey())) {
-                data = EncryptionUtils.encrypt(MobiComUserPreference.getInstance(context).getEncryptionKey(), data);
-            }
-            url = new URL(urlString);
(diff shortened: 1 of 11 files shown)


devashishmamgain (commented):
What are the methods that customer have to use

reytum:
Updated in the PR.

devashishmamgain:
@reytum can you share the sample code to call for updating metadata for a given groupID

reytum:
@devashishmamgain we do not have any methods exposed for groupId in Kommunicate android. Customer do not track the groupIds on mobile. We handle everything internally based on the use case. If some one needs To update via groupdId, we can give them the below method:

```java
 GroupInfoUpdate groupInfoUpdate = new GroupInfoUpdate(metadataForUpdate, groupId);

            KmUpdateConversationTask.KmConversationUpdateListener kmConversationUpdateListener = new KmUpdateConversationTask.KmConversationUpdateListener() {
                @Override
                public void onSuccess(Context context) {
                    if (callback != null) {
                        callback.onSuccess(channel, context);
                    }
                }

                @Override
                public void onFailure(Context context) {
                    if (callback != null) {
                        callback.onSuccess(channel, context);
                    }
                }
            };

            new KmUpdateConversationTask(context, groupInfoUpdate, kmConversationUpdateListener).executeOnExecutor(AsyncTask.THREAD_POOL_EXECUTOR);
```

kommunicate/src/main/java/com/applozic/mobicomkit/api/HttpRequestUtils.java
```diff
@@ -405,6 +405,55 @@ public String getResponse(String urlString, String contentType, String accept, b
         return null;
     }
 
+    public String makePatchRequest(String stringUrl, String data) throws Exception {
```
devashishmamgain, line 408: is it same code as other GET and POST call? can you move some common code to a separate common function, or may be pass the request method as a parameter if rest of the code is same

devashishmamgain (changes requested):

reytum:
Added new changes. Also added a function to update conversation:
```java
  GroupInfoUpdate groupInfoUpdate = new GroupInfoUpdate(metadataForUpdate, channel.getKey());

            KmSettings.updateConversation(context, groupInfoUpdate, new KmUpdateConversationTask.KmConversationUpdateListener() {
                @Override
                public void onSuccess(Context context) {
                    Utils.printLog(context, TAG, "Successfully updated conversation metadata");
                }

                @Override
                public void onFailure(Context context) {
                    Utils.printLog(context, TAG, "Failed to update conversation metadata for");
                }
            });
```

kommunicate/src/main/java/com/applozic/mobicomkit/api/HttpRequestUtils.java
reytum: Ok

reytum: Did the code cleanup. Added single point entry for request.

devashishmamgain (approved):

PR #204, opened 2021-06-14
create seperate methods for conversation update

reytum:
* Removed automatic updates from conversation builder
* Added methods to update teamId, conversationAssignee and conversationMetadata
* Moved conversation view to different method in detailed conversation adapter

### DIFF
--- a/app/src/main/assets/applozic-settings.json
+++ b/app/src/main/assets/applozic-settings.json
@@ -150,5 +150,9 @@
     "sendMessageOnSpeechEnd": true,
     "languageCode": "en-US"
   },
-  "hidePostCTA": false
+  "hidePostCTA": {
+    "submit": false,
+    "link": false,
+    "quickReply": false
+  }
 }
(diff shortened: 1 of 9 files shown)


app/src/main/assets/applozic-settings.json
```diff
@@ -150,5 +150,9 @@
     "sendMessageOnSpeechEnd": true,
     "languageCode": "en-US"
   },
-  "hidePostCTA": false
+  "hidePostCTA": {
```
devashishmamgain, line 153: What will happen to existing customers using old structure in json file and upgrades SDK without changing json file

devashishmamgain (changes requested):

app/src/main/assets/applozic-settings.json
reytum: Nothing will happen. This is not live yet.

devashishmamgain (approved):

PR #205, opened 2021-06-14
move receiver name outside and add setting to change the color

reytum:
* Moved the receiver name text view outside the message box
* Added setting to change the color of the receiver name text view

### DIFF
--- a/app/src/main/assets/applozic-settings.json
+++ b/app/src/main/assets/applozic-settings.json
@@ -154,5 +154,6 @@
     "submit": false,
     "link": false,
     "quickReply": false
-  }
+  },
+  "receiverNameTextColor": "#5C6677"
 }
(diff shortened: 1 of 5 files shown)


devashishmamgain:
@reytum can you share apk for this

app/src/main/assets/applozic-settings.json
```diff
@@ -154,5 +154,6 @@
     "submit": false,
     "link": false,
     "quickReply": false
-  }
+  },
+  "receiverNameTextColor": "#5C6677"
```
devashishmamgain, line 158: Is it receiver or sender? Name seems confusing

reytum: Its receiver. It can be agent or Bot for User.

devashishmamgain: If you send me a message then I m the receiver?

devashishmamgain: Left side one will be the sender right?

reytum: Yes. But I think it is easier as receiver as the word is being used in several other settings like receivedMessageTextColor, etc. Maybe we can make it as receivedMessageTitleColor.

devashishmamgain: receivedMessageTitleColor is better if we have something for message.
This is for name so we can go ahead with existing used

devashishmamgain (approved):

PR #209, opened 2021-06-25
release version V2.1.8

reytum:
* All commits for release version 2.1.8


local.properties
```diff
@@ -0,0 +1,10 @@
+## This file must *NOT* be checked into Version Control Systems,
+# as it contains information specific to your local configuration.
+#
+# Location of the SDK. This is only used by Gradle.
+# For customization when using a Version Control System, please read the
+# header note.
+#Fri Jun 25 14:18:43 IST 2021
+sdk.dir=/Users/reytum/Library/Android/sdk
```
devashishmamgain, line 8: is this required?

devashishmamgain (changes requested):

local.properties
reytum: No, But I committed the file because it had entries for jfrog. This path will be changed locally when someone clones the SDK.

devashishmamgain (approved):

PR #216, opened 2021-08-11
[ TD- 1828 ] Invalid conversation showing in Agent app fix

amntoppo:
- Conversations initiated by bots but had no reply should not show up in the Agent App
- Those conversations have Conversation Status = -1

### PR Checklist

- [x] I have tested it locally and all functionalities are working fine.
- [x] Deleted Shared Preferences and tested

### DIFF
--- a/kommunicate/src/main/java/com/applozic/mobicommons/people/channel/Channel.java
+++ b/kommunicate/src/main/java/com/applozic/mobicommons/people/channel/Channel.java
@@ -45,6 +45,7 @@ public class Channel extends JsonMarker {
     public static final int CLOSED_CONVERSATIONS = 3;
     public static final int ASSIGNED_CONVERSATIONS = 1;
     public static final int ALL_CONVERSATIONS = 2;
+    public static final int NOTSTARTED_CONVERSATIONS = -1;
     public static final String AL_BLOCK = "AL_BLOCK";
 
     public Channel() {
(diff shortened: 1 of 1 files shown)


kommunicate/src/main/java/com/applozic/mobicommons/people/channel/Channel.java
```diff
@@ -259,6 +260,10 @@ public int generateKmStatus(String loggedInUserId) {
             return CLOSED_CONVERSATIONS;
         }
 
+        if(getMetadata().containsKey(CONVERSATION_STATUS) && ("-1".equals(getMetadata().get(CONVERSATION_STATUS)))) {
+            return NOTSTARTED_CONVERSATIONS;
```
adarshmishra, line 264: @amntoppo how are we using this information to not showing initial conversation

amntoppo: On the Agent app, it fetches the conversations according to the Conversation status of the channel: Assigned, Resolved and All.
Earlier, these initial conversations were kept as ALL_CONVERSATIONS and so this was also getting fetched when agent app fetches for all conversations. Now that we have created a separate status of these conversations, it will be fetched only when required, like when we search for them.

adarshmishra (approved):

PR #228, opened 2021-09-30
Custom Chat Background (either color or local drawable) feature

Sathyan-Elangovan:
This Feature is to give the customisation for the conversation chat background.
1. Client can give custom color or can give drawable name after adding that drawable file to their app module.

### DIFF
--- a/kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/conversation/fragment/MobiComConversationFragment.java
+++ b/kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/conversation/fragment/MobiComConversationFragment.java
@@ -10,6 +10,7 @@ import android.content.Context;
 import android.content.DialogInterface;
 import android.content.Intent;
 import android.content.res.Configuration;
+import android.content.res.Resources;
 import android.database.Cursor;
 import android.graphics.Bitmap;
 import android.graphics.Color;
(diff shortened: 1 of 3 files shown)


kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/conversation/fragment/MobiComConversationFragment.java
```diff
@@ -442,9 +444,28 @@ protected Bitmap processBitmap(Object data) {
...
+
+            if (customChatBackground.contains("#") && customChatBackground.length() >= 7){
+                conversationRootLayout.setBackgroundColor(Color.parseColor(alCustomizationSettings.getChatBackgroundColorOrDrawable()));
+            }else{
+                Resources resources = getResources();
+                int resourceId = resources.getIdentifier(customChatBackground, "drawable",getContext().getPackageName());
+                if (resourceId != 0){
+                    conversationRootLayout.setBackgroundResource(resourceId);
+                }
+            }
```
amntoppo, line 460: If someone adds an invalid drawable or an invalid color, it should give an error message and set the default background color.

Sathyan-Elangovan: @amntoppo The Above changes will set the default background only(If invalid input comes). But Do we really need to show the error in Toast?

amntoppo: If default color will be set if invalid resource is coming, then it should be fine.

amntoppo (approved):

PR #234, opened 2021-10-18
[TD-1865] - Collect email ID from user for Away Message for anonymous users

amntoppo:
- https://kommunicate.atlassian.net/wiki/spaces/DD/pages/940310533/Away+message+in+Android+SDK
- Created a Away View layout inside messagelist layout
- Separated AwayView class from ConversationFragment for clean code
- Created a separate updateEmail method using same AsyncTask to update anonymous user's email with parameters "elasticUpdate=true" and "allowEmail=true"


amntoppo:
It's a small conflict, I will resolve it when PR is approved. If I resolve now, it will be hard to review with all the previous code.

PR #239, opened 2021-10-28
[CM-698] - Update Away/Online status using Mqtt

amntoppo:
- Away/online status changed is published to a topic: https://kommunicate.atlassian.net/browse/TD-1641
- Message type is "APPLOZIC_25" - fetches for it, then sends a Broadcast
- Updated the UI when we receive the away/online status
- Away message is stored in cache and API away is only called once.

### DIFF
--- a/kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/conversation/MobiComKitBroadcastReceiver.java
+++ b/kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/conversation/MobiComKitBroadcastReceiver.java
@@ -4,6 +4,8 @@ import android.content.BroadcastReceiver;
 import android.content.Context;
 import android.content.Intent;
 import androidx.fragment.app.FragmentActivity;
+import io.kommunicate.utils.KmConstants;
+
 import android.text.TextUtils;
 
 import com.applozic.mobicomkit.ApplozicClient;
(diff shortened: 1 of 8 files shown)


kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/conversation/MobiComKitBroadcastReceiver.java
```diff
@@ -125,6 +125,8 @@ public void onReceive(Context context, Intent intent) {
             conversationUIService.updateMessageMetadata(keyString);
         } else if (BroadcastService.INTENT_ACTIONS.MUTE_USER_CHAT.toString().equals(action)) {
             conversationUIService.muteUserChat(intent.getBooleanExtra("mute", false), intent.getStringExtra("userId"));
+        } else if(BroadcastService.INTENT_ACTIONS.AGENT_STATUS.toString().equals(action)) {
+            conversationUIService.updateAgentStatus(intent.getStringExtra("userId"), intent.getIntExtra("status", 3));
```
Sathyan-Elangovan, line 129: here 3 is agent status online right?If yes Can you use `STATUS_ONLINE` to avoid confusion in future

Sathyan-Elangovan (changes requested):

kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/conversation/MobiComKitBroadcastReceiver.java
amntoppo: @Sathyan-Elangovan I have updated it

Sathyan-Elangovan (approved):

mukeshthawani (approved):

PR #246, opened 2021-12-14
V2.2.6 - Release to support Android 12

amntoppo:
Android 12 requires Activity/Services using intent filters to declare if they are exported (can be opened from outside the application) so did those changes and targeted API 31


Sathyan-Elangovan (approved):

kommunicate/build.gradle
```diff
@@ -8,17 +8,17 @@ rootProject.allprojects {
...
         abortOnError false
     }
 
     defaultConfig {
         minSdkVersion 16
-        targetSdkVersion 30
+        targetSdkVersion 31
         versionCode 1
-        versionName "2.2.5.1"
+        versionName "2.2.6"
```
mukeshthawani, line 21: Should we change the version to 2.3.0 because of a major change(SDK version update)?

amntoppo: Okay we should do that

amntoppo:
Using v2.3.0 for Android 12 support

PR #247, opened 2021-12-14
[CM-753] - Use GCM encryption

amntoppo:
> From Google: Google recommends that developers use "AES/GCM/NoPadding" instead of the aforementioned insecure mode.

Changed encryption type from "AES/ECB/NoPadding" to "AES/GCM/NoPadding".


mukeshthawani (approved):

amntoppo:
Requires server change. Will reopen once server changes are done and tested.

PR #252, opened 2021-12-20
[CM-726] - new Timestamp Design

amntoppo:
- Added new Timestamp design to replicate chat widget and use same timestamp as we show for Rich Message. (outside the message box)
- Added a boolean in applozic-settings "innerTimestampDesign" which when set to true will use the old design. If customers want to use old version.
- set default color of timestamp for both the design. If customers want to overwrite it, they can change it from applozic-settings: "sentMessageCreatedAtTimeColor" and "receivedMessageCreatedAtTimeColor"
- Added an onClickListener to messages which will show the timestamp (if grouped and hidden) when clicked.

Screenshot_20211220-174806
Screenshot_20211220-174820


Sathyan-Elangovan (approved):

kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/conversation/adapter/DetailedConversationAdapter.java
```diff
@@ -280,11 +291,21 @@ protected Bitmap processBitmap(Object data) {
             View v6 = layoutInflater.inflate(R.layout.km_feedback_agent_layout, parent, false);
             return new MyViewHolder6(v6);
         } else if (viewType == 0) {
-            View v0 = layoutInflater.inflate(R.layout.mobicom_received_message_list_view, parent, false);
+            View v0;
+            if(useInnerTimeStampDesign) {
+                v0 = layoutInflater.inflate(R.layout.mobicom_received_message_list_view, parent, false);
```
adarshmishra, line 296: @amntoppo rename variable to something meaningful, like recerverView etc

amntoppo: @adarshmishra I have changed name to view0, view 1. Also, we have 6 different view holders so I have commented out the role of each of them.

adarshmishra (approved):

PR #269, opened 2022-05-02
Support for Localization of editTextHintText

amntoppo:
`messageEditText.setHint(!TextUtils.isEmpty(alCustomizationSettings.getEditTextHintText()) ? alCustomizationSettings.getEditTextHintText() : getString(R.string.enter_message_hint));`

editTextHintText checked for value in alCustomizationSettings (applozic-settings.json) before localization using strings. 
Inside alCustomizationSettings, default value was set so localization using string was not happening.
Removed the default value from alCustomizationSettings  so now localization as well as custom settings will work.

### DIFF
--- a/app/src/main/assets/applozic-settings.json
+++ b/app/src/main/assets/applozic-settings.json
@@ -119,7 +119,7 @@
     ":file": true,
     ":audio": true
   },
-  "editTextHintText": "Type your message...",
+  "editTextHintText": "",
   "replyMessageLayoutSentMessageBackground": "#C0C0C0",
   "replyMessageLayoutReceivedMessageBackground": "#F5F5F5",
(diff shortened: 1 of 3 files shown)


Sathyan-Elangovan (changes requested):
What Will happen if customer adds a string for "editTextHintText" in Applozic-settings.json file?

amntoppo:
@Sathyan-Elangovan 
If a customer adds a string to "editTextHintText" : then that string will be shown, irrespective of what the localization string is.
If a customer adds a string to both "editTextHintText" and R.string.enter_message_hint: "editTextHintText" will be shown
If a customer adds a string to R.string.enter_message_hint: then this text will be shown

Sathyan-Elangovan (approved):

PR #270, opened 2022-05-09
Patch release - V2.3.9

amntoppo:
- Added EditTextHintText localization
- Fix logging to txt file

### DIFF
--- a/kommunicate/src/main/java/com/applozic/mobicommons/commons/core/utils/Utils.java
+++ b/kommunicate/src/main/java/com/applozic/mobicommons/commons/core/utils/Utils.java
@@ -339,7 +339,7 @@ public class Utils {
             BufferedWriter bufferedWriter = null;
             try {
                 String folder = "/" + Utils.getMetaDataValue(context, "main_folder_name");
-                File dir = new File(Environment.getExternalStorageDirectory().getAbsolutePath() + folder);
+                File dir = new File(context.getFilesDir().getAbsolutePath() + folder);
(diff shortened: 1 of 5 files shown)


kommunicate/src/main/java/com/applozic/mobicommons/commons/core/utils/Utils.java
```diff
@@ -339,7 +339,7 @@ public static void writeToFile(Context context, String log) {
             BufferedWriter bufferedWriter = null;
             try {
                 String folder = "/" + Utils.getMetaDataValue(context, "main_folder_name");
-                File dir = new File(Environment.getExternalStorageDirectory().getAbsolutePath() + folder);
+                File dir = new File(context.getFilesDir().getAbsolutePath() + folder);
```
Sathyan-Elangovan, line 342: Why we are changing  from `Environment.getExternalStorageDirectory()` to `context.getFilesDir()`

amntoppo: `Environment.getExternalStorageDirectory()` is deprecated and will fail if external SD card is not present. 
Using `context.getFilesDir() `we will be logging to the correct directory of the application.

amntoppo: image

Sathyan-Elangovan (approved):

PR #273, opened 2022-05-18
[CM-923] - Notify everybody notification for Agent app

amntoppo:
- Added support to show notification on Agent app when "Notify everybody" is selected from dashboard.


Sathyan-Elangovan (approved):

kommunicate/src/main/java/com/applozic/mobicomkit/api/account/user/MobiComUserPreference.java
```diff
@@ -968,4 +969,17 @@ public void setLoggedUserDeletedFromDashboard(boolean deletedFromDashboard) {
...
     }
+
+    public boolean isAgentRouting() {
+        if(sharedPreferences != null) {
+            return sharedPreferences.getBoolean(AGENT_ROUTING, true);
+        }
+        return false;
+    }
+
+    public void setAgentRouting(boolean isAgentRouting) {
```
mukeshthawani, line 980: Should we change this to `notifyEverybody`? If that's what we are doing then it will be better to set the same here and in other methods.

mukeshthawani (approved):

PR #279, opened 2022-06-17
[CM-951] - add support for pseudo name

amntoppo:
## What do you want to achieve?
- Pseudo names are fetched from App Settings API, "userName".
- When we create a random userID for visitors, fetch "userName" from API and set it as display Name of user
- Removed duplicate code - KmGetAppListTask which is same as KmAppSettingTask


kommunicate/src/main/java/io/kommunicate/KmConversationHelper.java
```diff
@@ -694,10 +694,11 @@ private static KmCallback getCallbackWithAppSettingsToCreateConversation(final b
         return new KmCallback() {
             @Override
             public void onSuccess(Object message) {
-                KmAppSettingModel.KmResponse kmAppSettings = (KmAppSettingModel.KmResponse) message;
+                KmAppSettingModel kmAppSettings = (KmAppSettingModel) message;
                 if (kmAppSettings != null) {
                     List<String> agents = new ArrayList<>();
-                    agents.add(kmAppSettings.getAgentId());
+                    agents.add(kmAppSettings.getResponse().getAgentId());
+                    //conversationBuilder.getKmUser().setDisplayName(kmAppSettings.getUserName());
```
Sathyan-Elangovan, line 701: remove commented codes

kommunicate/src/main/java/io/kommunicate/Kommunicate.java
```diff
@@ -802,8 +801,25 @@ private static String getClientGroupId(String userId, List<String> agentIds, Lis
...
         user.setUserId(generateUserId());
+        new KmAppSettingTask(ApplozicService.getAppContext(),
+                MobiComKitClientService.getApplicationKey(ApplozicService.getAppContext()),
+                new KmCallback() {
+                    @Override
+                    public void onSuccess(Object message) {
+                        final KmAppSettingModel appSettingModel = (KmAppSettingModel) message;
+                        if (appSettingModel != null && appSettingModel.getResponse() != null && appSettingModel.getChatWidget() != null) {
+                            if (appSettingModel.getChatWidget().isPseudonymsEnabled()) {
+                                user.setDisplayName(appSettingModel.getResponse().getUserName());
```
Sathyan-Elangovan, line 814: what if `appSettingModel.getResponse().getUserName()` is nil?

Sathyan-Elangovan (changes requested):

kommunicate/src/main/java/io/kommunicate/Kommunicate.java
amntoppo: It should not be null if getChatWidget().isPseudonymsEnabled() is true, but I have added a TextUtils.isEmpty check for safety.

Sathyan-Elangovan (approved):

kommunicate/src/main/java/io/kommunicate/KmConversationHelper.java
```diff
@@ -694,10 +694,10 @@ private static KmCallback getCallbackWithAppSettingsToCreateConversation(final b
         return new KmCallback() {
             @Override
             public void onSuccess(Object message) {
-                KmAppSettingModel.KmResponse kmAppSettings = (KmAppSettingModel.KmResponse) message;
+                KmAppSettingModel kmAppSettings = (KmAppSettingModel) message;
                 if (kmAppSettings != null) {
                     List<String> agents = new ArrayList<>();
-                    agents.add(kmAppSettings.getAgentId());
+                    agents.add(kmAppSettings.getResponse().getAgentId());
```
adarshmishra, line 700: @amntoppo How are we handling error here, will it create any issue if response from server is not proper or call failed?

amntoppo: @adarshmishra it will show "Unable to create conversation" for this scenario. For Pseudo name, it will create a user without the pseudoname.

adarshmishra (approved):

PR #286, opened 2022-07-13
[CM-975] Added a link drawable in link buttons

pahadi777:
Added a link drawable for every rich messages that contains link . It was available in the web already but now implemented it in Android too .

IMG_20220713_183859


kommunicate/src/main/java/io/kommunicate/utils/KmUtils.java
```diff
@@ -75,6 +73,13 @@ public static void setGradientStrokeColor(View view, int width, int color) {
         gradientDrawable.setStroke(width, color);
     }
 
+    public static void setDrawable(TextView textView, int drawableRes, int color)
```
amntoppo, line 76: Method name "setDrawable" does not specify what exactly it does, name it in a way which tells what exactly this method does, so it can be used later on in other places.

amntoppo (changes requested):

amntoppo (approved):

Sathyan-Elangovan (approved):
Add a changelog entry

PR #291, opened 2022-08-02
[CM-1038] Crash Fix while calling getName() at KmDocumentView.java:352

pahadi777:
Added a null check at KMDocumentView.java : 352 which removes the crashing

### DIFF
--- a/kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/attachmentview/KmDocumentView.java
+++ b/kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/attachmentview/KmDocumentView.java
@@ -350,7 +350,7 @@ public class KmDocumentView {
 
     public void playAudio() {
         final String mimeType;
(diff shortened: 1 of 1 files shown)


kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/attachmentview/KmDocumentView.java
```diff
@@ -349,7 +349,13 @@ public void onClick(View v) {
     }
 
     public void playAudio() {
-        final String mimeType = FileUtils.getMimeType(message.getFileMetas().getName());
+        final String mimeType;
+        if(message.getFileMetas() != null) {
```
amntoppo, line 353: FileMeta can be non-null but getName() can still be null, it will throw an exception again

amntoppo (changes requested):

Sathyan-Elangovan (approved):

amntoppo (approved):

PR #294, opened 2022-08-04
[CM-1036] Added a default last message body when last message of a conversation has an empty body

pahadi777:
Earlier ->

Screenshot 2022-07-29 at 12 21 07 PM-20220729-065113 (1)

Now ->

IMG_20220804_122528

### DIFF
--- a/kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/conversation/adapter/QuickConversationAdapter.java
+++ b/kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/conversation/adapter/QuickConversationAdapter.java
@@ -3,6 +3,7 @@ package com.applozic.mobicomkit.uiwidgets.conversation.adapter;
 import android.app.Activity;
 import android.content.Context;
 import android.graphics.Bitmap;
+import android.graphics.Color;
 import android.graphics.drawable.GradientDrawable;
 
 import androidx.fragment.app.FragmentActivity;
@@ -58,6 +59,7 @@ import java.util.Locale;
 import java.util.Map;
 
 import de.hdodenhof.circleimageview.CircleImageView;
+import io.kommunicate.utils.KmUtils;
 
 /**
  * Created by adarsh on 4/7/15.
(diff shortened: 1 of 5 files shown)


kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/conversation/adapter/QuickConversationAdapter.java
```diff
@@ -250,6 +251,14 @@ public void onBindViewHolder(RecyclerView.ViewHolder holder, int position) {
                     showConversationSourceIcon(channel, myholder.attachmentIcon);
                 }
 
+                if(TextUtils.isEmpty(myholder.messageTextView.getText()))
+                {
+                    myholder.messageTextView.setCompoundDrawablesWithIntrinsicBounds(R.drawable.ic_messageicon, 0, 0, 0);
+                    myholder.messageTextView.setCompoundDrawablePadding(20);
+                    myholder.messageTextView.setText(DEFAULT_MSG_BODY);
+                }
+                else myholder.messageTextView.setCompoundDrawablesWithIntrinsicBounds(0, 0, 0, 0);
```
Sathyan-Elangovan, line 260: Have checked by creating more number of conversations,,so that you can scroll the recycler view & check this change is working fine or not

Sathyan-Elangovan (commented):
I guess ticket mentioned as  CM-1038 is wrong. There is one more PR with same ticket is there.

kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/conversation/adapter/QuickConversationAdapter.java
pahadi777: Ya ya the else part is handling that thing . I have tested it with decent number of conversations .

Sathyan-Elangovan (approved):

kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/conversation/adapter/QuickConversationAdapter.java
```diff
@@ -250,6 +251,14 @@ public void onBindViewHolder(RecyclerView.ViewHolder holder, int position) {
                     showConversationSourceIcon(channel, myholder.attachmentIcon);
                 }
 
+                if(TextUtils.isEmpty(myholder.messageTextView.getText()))
+                {
```
amntoppo, line 255: This check is not correct, it might fail for some slow Android devices, during repeated scrolling.
Rather than checking if messageTextView is empty (which we set empty in previous lines). Add a else if statement which checks if the message is a Rich message.
Message.isRichMessage() should tell you that.

amntoppo (changes requested):

kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/conversation/adapter/QuickConversationAdapter.java
amntoppo, line 260: Move this method to a different class, which can be used again.

amntoppo (approved):

PR #295, opened 2022-08-05
[CM-1037] - fix for attachment not sent

amntoppo:
## Issue:
- Happens only with slow internet: If you send multiple attachments at once, sometimes server gives 502 response code. This response code was not handled by device and message was being sent as empty message.
- Empty message in dashboard:
image

## Fix:
- Exception will be caught when any other response comes, other than 200. This will be handled by showing retry option in attachment. When you click retry, attachments are sent.
- Retry option: 
image

### DIFF
--- a/kommunicate/src/main/java/com/applozic/mobicomkit/api/conversation/MessageClientService.java
+++ b/kommunicate/src/main/java/com/applozic/mobicomkit/api/conversation/MessageClientService.java
@@ -410,7 +410,7 @@ public class MessageClientService extends MobiComKitClientService {
             for (String filePath : message.getFilePaths()) {
                 try {
                     String fileMetaResponse = new FileClientService(context).uploadBlobImage(filePath, handler, oldMessageKey);
-                    if (fileMetaResponse == null) {
+                    if (TextUtils.isEmpty(fileMetaResponse)) {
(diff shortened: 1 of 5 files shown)


kommunicate/src/main/java/com/applozic/mobicomkit/api/conversation/MessageClientService.java
```diff
@@ -426,8 +426,7 @@ public void processMessage(Message message, Handler handler, String userDisplayN
                         }
                         BroadcastService.sendMessageUpdateBroadcast(context, BroadcastService.INTENT_ACTIONS.UPLOAD_ATTACHMENT_FAILED.toString(), message);
                         return;
-                    }
-                        if (!TextUtils.isEmpty(fileMetaResponse)) {
+                    } else if (!TextUtils.isEmpty(fileMetaResponse)) {
```
Sathyan-Elangovan, line 429: Do we need `else if` here for this `!TextUtils.isEmpty(fileMetaResponse)`

Sathyan-Elangovan (commented):
Add a video or screen shot for reference please

Sathyan-Elangovan (approved):

PR #297, opened 2022-08-19
[CM-1063] - Upgrade channel list API to v5

amntoppo:
List v3 API provides a lot of unnecessary data which are not required, because of which we are getting outOfMemoryError. 

To fix this, upgrade the list api to v5.

Schema is same, only difference is that v5 API does not give group member details for every group


Sathyan-Elangovan (approved):
@amntoppo Add an entry in Changelog

PR #300, opened 2022-08-25
info option long press fix new

rajeev-kommunicate:
info option long press fix new


Sathyan-Elangovan:
@rajeev-kommunicate what happened to this PR?if we don't need this..we can close it.

rajeev-kommunicate:
On backend this was intentionally disabled.

PR #307, opened 2022-09-12
[CM-1094] - SDK Optimization 

amntoppo:
## Issue: 
Getting **Application Not Responding** when local database is big in size

## Root cause:
- We were using "synchronized" method calls at some places where it was not required. Using synchronized method calls block the other thread while the current thread is doing the work in that class. It was also blocking the main thread at some places.
- Database Input / Output were done in main thread. It was causing performance issue.

## Fix:
- Removed "synchronized" method calls wherever it's not necessary. We do not need synchronized calls as we are already sending a Broadcast to update the UI. With this, no other threads will be blocked, and UI will also be updated whenever necessary.
- Added Channel User Mapper to Channel Data class. With this, we do not need to access the Database everytime to get channel user details, we can fetch it from memory which is much faster.
- Moved time-consuming Database Input/Output process to background thread.
- Removed API call which were unnecessary - "rest/ws/user/chat/mute/list"

### DIFF
--- a/kommunicate/src/main/java/com/applozic/mobicomkit/api/conversation/database/MessageDatabaseService.java
+++ b/kommunicate/src/main/java/com/applozic/mobicomkit/api/conversation/database/MessageDatabaseService.java
@@ -477,7 +477,7 @@ public class MessageDatabaseService {
         return empty;
     }
 
-    public synchronized void updateMessageFileMetas(long messageId, final Message message) {
+    public void updateMessageFileMetas(long messageId, final Message message) {
         ContentValues values = new ContentValues();
         values.put("keyString", message.getKeyString());
         if (message.getFileMetaKeyStrings() != null) {
@@ -500,7 +500,7 @@ public class MessageDatabaseService {
         dbHelper.close();
     }
 
-    public synchronized long createMessage(final Message message) {
+    public long createMessage(final Message message) {
         long id = -1;
         if (message.getMessageId() != null) {
             return message.getMessageId();
@@ -521,7 +521,7 @@ public class MessageDatabaseService {
         return id;
     }
 
-    public synchronized long createSingleMessage(final Message message) {
+    public long createSingleMessage(final Message message) {
         SQLiteDatabase database = dbHelper.getWritableDatabase();
         ApplozicClient applozicClient = ApplozicClient.getInstance(context);
         long id = -1;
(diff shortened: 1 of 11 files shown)


kommunicate/src/main/java/com/applozic/mobicomkit/api/conversation/database/MessageDatabaseService.java
```diff
@@ -1206,7 +1206,7 @@ public void deleteChannelConversation(Integer channelKey) {
         Utils.printLog(context, TAG, "Delete " + deletedRows + " messages.");
     }
 
-    public synchronized void updateContactUnreadCount(String userId) {
+    public void updateContactUnreadCount(String userId) {
         try {
```
adarshmishra, line 1210: Removing synchronized might create issues, as it is reading values and incrementing. Two threads might be reading the same value and updating the count by 1.

kommunicate/src/main/java/com/applozic/mobicomkit/channel/service/ChannelService.java
```diff
@@ -243,7 +243,7 @@ public List<Channel> getChannelList() {
         return channelDatabaseService.getAllChannels();
     }
 
-    public synchronized void syncChannels(boolean isMetadataUpdate) {
+    public void syncChannels(boolean isMetadataUpdate) {
         try {
             final MobiComUserPreference userpref = MobiComUserPreference.getInstance(context);
```
adarshmishra, line 248: Check this one also, it might create issues as we are updating time. Also, this might leads to too many calls in a small amount of time.

kommunicate/src/main/java/com/applozic/mobicomkit/api/conversation/database/MessageDatabaseService.java
amntoppo: As per the [documentation](http://developer.android.com/reference/android/database/sqlite/SQLiteDatabase.html#setLockingEnabled(boolean)), access to the SQLiteDatabase is synchronized by default, thus guaranteeing that no two threads will ever touch it at the same time.

amntoppo: We do not need to synchronize manually

kommunicate/src/main/java/com/applozic/mobicomkit/channel/service/ChannelService.java
amntoppo: The issue with this was: Because of this syncChannels() method, other synchronized methods were getting thread locked, because other methods were also synchronized.
Now, I have made this method syncChannels() synchronized, but removed other method's synchronization. This should work because we are sending a Broadcast whenever we find that there is an update in Channel.

adarshmishra (approved):
@surajkukkipady18 these changes need proper testing. Make sure to cover scenario like internet offline , online , syncing older messages properly etc

Sathyan-Elangovan (approved):

PR #336, opened 2023-01-06
Minor fix for Zendesk

amntoppo:
## Issue:
1. Transcript messaging sending "Admin created Group Conversation"
2. Listening to chat events even before handoff (causing crash sometimes)

### DIFF
--- a/kommunicate/src/main/java/io/kommunicate/zendesk/KmZendeskClient.java
+++ b/kommunicate/src/main/java/io/kommunicate/zendesk/KmZendeskClient.java
@@ -309,7 +309,11 @@ private void processAgentLeave() {
                     listOfMessage = Arrays.asList(kmConversationResponse.getMessage());
                     Collections.reverse(listOfMessage);
                     for(Message message : listOfMessage) {
-                        String username;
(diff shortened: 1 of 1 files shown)


kommunicate/src/main/java/io/kommunicate/zendesk/KmZendeskClient.java
```diff
@@ -305,7 +309,11 @@ public void run() {
                     listOfMessage = Arrays.asList(kmConversationResponse.getMessage());
                     Collections.reverse(listOfMessage);
                     for(Message message : listOfMessage) {
-                        String username;
+                        if (Message.GroupMessageMetaData.FALSE.getValue().equals(message.getMetaDataValueForKey(Message.GroupMessageMetaData.KEY.getValue()))) {
+                            continue;
```
adarshmishra, line 313: @amntoppo what is this check?

amntoppoKM:
```
        KEY("show"),
        HIDE_KEY("hide"),
        FALSE("false"),
        TRUE("true");
```
The first message contains `"show": "false"`

We are checking if the message contains show: false

adarshmishra (approved):

amntoppoKM (approved):

PR #340, opened 2023-01-19
[CM-1275] GeneralException: Application key or UserId as request body or device key as header not passed

pahadi777:
Stopped calling refresh token API whenever Application key or UserId as request body or device key as header not passed


kommunicate/src/main/java/com/applozic/mobicomkit/api/account/register/RegisterUserClientService.java
```diff
@@ -202,6 +202,10 @@ public RegistrationResponse createAccount(User user) throws Exception {
     }
 
     public boolean refreshAuthToken(String applicationId, String userId) {
+        MobiComUserPreference mobiComUserPreference = MobiComUserPreference.getInstance(context);
```
Sathyan-KM, line 205: @pahadi777 Did you test the api by passing one parameter ?As of now we have 
1. applicationid
2. userid
3. device key
Just test by passing only one param for  each & check what's happening

pahadi777: Passing any parameter as null is giving failure on calling refreshAuthToken . Only passing all the parameters is giving success .

Sathyan-KM (approved):
Add changelog entry

adarshmishra (approved):

PR #341, opened 2023-01-19
[CM-1261] - Handle notifications with FCM payload

amntoppoKM:
## Summary
- Whenever we received a FCM notification, we were not using the FCM payload to show notification. Rather, we were calling the sync message API which was called in background and notifications were shown with it.
- This was causing problems as Android OS sometimes does not allow background tasks to be executed to save memory. Due to this, notifications were not shown sometimes.

# Fix:
- Now, we are using FCM payload to directly show the notification, without calling the sync call API. 
- Sync call API will be called when we click the notification or when the app is opened, which will ensure that no messages are lost.
- We are calling "message/delivered" API in the background to send delivery report. Even if this API is not called by Android OS, delivery report will be sent when user clicks on the notification or when app is opened.
- We are using the FCM payload to also update the channel, which is required by Agent app to decide which notification to be shown.

### DIFF
--- a/kommunicate/src/main/java/com/applozic/mobicomkit/api/conversation/MobiComMessageService.java
+++ b/kommunicate/src/main/java/com/applozic/mobicomkit/api/conversation/MobiComMessageService.java
@@ -7,6 +7,7 @@ import android.net.Uri;
 import androidx.localbroadcastmanager.content.LocalBroadcastManager;
 
 import android.text.TextUtils;
+
 import com.applozic.mobicomkit.ApplozicClient;
 import com.applozic.mobicomkit.api.MobiComKitConstants;
 import com.applozic.mobicomkit.api.account.user.MobiComUserPreference;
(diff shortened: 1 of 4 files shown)


kommunicate/src/main/java/com/applozic/mobicomkit/api/conversation/MobiComMessageService.java
```diff
@@ -457,4 +458,47 @@ public synchronized void processInstantMessage(Message message) {
...
+        Channel channel = null;
+        if (message.getGroupId() != null) {
+            channel = ChannelService.getInstance(context).getChannelInfo(message.getGroupId());
+        }
+        try {
+            List<String> messageKeys = new ArrayList<>();
+            if (message.getMetadata() != null && message.getMetaDataValueForKey(Message.MetaDataType.AL_REPLY.getValue()) != null && !messageDatabaseService.isMessagePresent(message.getMetaDataValueForKey(Message.MetaDataType.AL_REPLY.getValue()))) {
+                messageKeys.add(message.getMetaDataValueForKey(Message.MetaDataType.AL_REPLY.getValue()));
+            }
+            if (messageKeys != null && messageKeys.size() > 0) {
```
adarshmishra, line 473: messageKeys is already defined, remove unnecessary code.

adarshmishra (approved):

PR #342, opened 2023-01-19
[CM-1117] Added platform flag for User

pahadi777:
Added platform flag in the Android SDK


kommunicate/src/main/java/com/applozic/mobicomkit/api/account/user/User.java
```diff
@@ -17,6 +17,7 @@
 public class User extends JsonMarker {
 
     private static final String DEFAULT_USER_ID_REGEX = "^[a-zA-Z0-9_+#@.?|=;-]+$";
+    private static int platformFlag = 1;
```
Sathyan-KM, line 20: Use `platform`

Sathyan-KM: create enum for other platform values

pahadi777: Changed as suggested

Sathyan-KM (approved):
Add description to the PR. Add changelog

PR #349, opened 2023-02-03
[CM-1309] Fixed agent app toolbar showing conversation assignee image instead of end user image

pahadi777:
Previously inside the `updateSupportGroupTitleAndImageAndHideSubtitle` function there was no separate case for agent app so even when the toolbar was of agent app it was working as a toolbar of SDK which is fixed now


amntoppoKM:
@pahadi777 When agent's image is not available, what will happen?

pahadi777:
> 

It will skip the `if` and `else if` part and go inside the `else` part where the alphabetic image will be shown .

pahadi777:
> > 
> 
> It will skip the `if` and `else if` part and go inside the `else` part where the alphabetic image will be shown .

Alphabetic image shown will be of the user

amntoppoKM (approved):

PR #365, opened 2023-03-03
[CM-1353] - Added support to delete conversation for end-user

amntoppoKM:
## Summary:
- To delete a conversation, a conversation should be long pressed. A "Delete conversation" option will show.
- API call used: `"rest/ws/group/delete"`
- By default, delete option is disabled. to enable, add `"deleteOption": true` in kommunicate-settings.json file
- Conversation will be deleted from both end-user and dashboard

## Screenshot:
image


Sathyan-KM:
@amntoppo resolve the conflicts . There are older changes present in the PR. I guess resolving conflicts would remove these  unnecessary changes

Sathyan-KM (approved):

PR #391, opened 2023-05-30
[CM-1412] - Agent app ANR - KmMessageProperties.setUnreadCount

pranaysaini04:
## Summary

- Fixed slow io operation in Main thread caused by getUnreadMessageCountForChannel method in MessageDatabaseService class


amntoppoKM:
@pranaysaini04 Don't create the thread in Database service. Create the thread where the database service is called from

PR #393, opened 2023-06-01
[CM-1377]- Add support for Custom Input Field

amntoppoKM:
## What do you want to achieve?
- Added support for Custom Input FIeld: https://docs.kommunicate.io/docs/message-types#custom-input-field
- Validation and Action are optional. 
- Validation regex differs in Javascript and Java. Same regex was giving PatternSyntaxException in Java, so added default regex for email and phone in case PatternSyntaxException is thrown.
- When regex does not match, show error toast message from - "errorText" -> If errorText not present then "Invalid validation regex" toast message
- If replyMetadata is present, then send the message message by adding this metadata to trigger next intent

### PR Checklist

- [x] I have tested it locally and all functionalities are working fine.
- [x] I have compared it with mocks and all design elements are the same.

### How was the code tested?

- 
`{
    "message": "",
    "platform": "kommunicate",
    "metadata": {
        "templateId": "13",
        "KM_FIELD": {
            "label": "Email",
            "field": "email",
            "fieldType": "EMAIL",
            "placeholder": "Enter your email",
            "action": {
                "updateUserDetails": true
            },
            "validation": {
                "regex": "^(([^<>()\\[\\]\\.;:\\s@\"]+(\\.[^<>()[\\]\\.,;:\\s@\"]+)*)|(\".+\"))@((\\[[0-9]{1,3}\\.[0-9]{1,3}\\.[0-9]{1,3}\\.[0-9]{1,3}\\])|(([a-zA-Z\\-0-9]+\\.)+[a-zA-Z]{2,}))$",
                "errorText": "The email you have entered is either invalid or not accepted"
            }
        },
        "replyMetadata": {
            "KM_TRIGGER_EVENT": "phone number"
        }
    }
}`

##Screenshot
image
image

### DIFF
--- a/kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/conversation/fragment/MobiComConversationFragment.java
+++ b/kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/conversation/fragment/MobiComConversationFragment.java
@@ -4242,15 +4242,15 @@ public abstract class MobiComConversationFragment extends Fragment implements Vi
         switch(fieldType) {
             case(KmCustomInputModel.EMAIL) : {
                 user.setEmail(message);
-                return;
+                break;
             }
             case(KmCustomInputModel.NAME) : {
                 user.setDisplayName(message);
-                return;
+                break;
             }
             case(KmCustomInputModel.PHONE_NUMBER) : {
                 user.setContactNumber(message);
-                return;
+                break;
             }
             default: {
                 Map<String, String> inputMetadata = new HashMap<>();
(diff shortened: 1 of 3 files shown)


kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/conversation/fragment/MobiComConversationFragment.java
```diff
@@ -1232,9 +1241,6 @@ protected void sendMessage() {
                 processSendMessage();
             }
         }
-        if (contact != null && !contact.isBlocked() || channel != null) {
-            handleSendAndRecordButtonView(false);
-        }
     }
```
Sathyan-KM, line 1244: removing this from here won't affect normal flow ?

Sathyan-KM (changes requested):

kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/conversation/fragment/MobiComConversationFragment.java
amntoppoKM: No, because we have added handleSendAndRecordButtonView were already present in other places and I have tested and added handleSendAndRecordButtonView wherever it was required.

Sathyan-KM (approved):

PR #428, opened 2023-08-17
[CM-1595] Fixed toolbar title alignement when menu icon is disabled

himanshujoshi848:
## When menu icon is enabled
Screenshot_20230816-175544

## When menu icon is disabled

Screenshot_20230816-175559


pranaysaini04 (approved):

Sathyan-KM:
@himanshujoshi848 whenever you add screenshot please mention it clearly that whether its after fix or before fix and write the summary clearly. The issue may be clear to you not for the Reviewer or others.They will get to know from the PR only.

PR #452, opened 2023-11-14
[CM-1734] Fixed unread count returing wrong value

himanshujoshi848:
## Summary

- Unread count from sever was returning a wrong value which is fixed.

## Testing

- Use the below code in `SplashScreenActivity.java` to fetch unread count from the server ->

```
    new AlTotalUnreadCountTask(this, new AlTotalUnreadCountTask.TaskListener() {
        @Override
        public void onSuccess(Integer unreadCount) {
            Log.d("serverCount",String.valueOf(unreadCount));
        }
        @Override
        public void onFailure(String error) {

        }
    }).execute();
```

### DIFF
--- a/kommunicate/src/main/java/com/applozic/mobicomkit/api/conversation/AlTotalUnreadCountTask.java
+++ b/kommunicate/src/main/java/com/applozic/mobicomkit/api/conversation/AlTotalUnreadCountTask.java
@@ -5,9 +5,21 @@ import android.content.Context;
 import com.applozic.mobicomkit.ApplozicClient;
 import com.applozic.mobicomkit.api.account.user.MobiComUserPreference;
 import com.applozic.mobicomkit.api.conversation.database.MessageDatabaseService;
+import com.applozic.mobicomkit.broadcast.BroadcastService;
+import com.applozic.mobicomkit.channel.service.ChannelService;
+import com.applozic.mobicomkit.feed.ChannelFeed;
 import com.applozic.mobicommons.ApplozicService;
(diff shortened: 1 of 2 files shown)


kommunicate/src/main/java/com/applozic/mobicomkit/api/conversation/AlTotalUnreadCountTask.java
```diff
@@ -27,12 +39,16 @@ public AlTotalUnreadCountTask(Context context, TaskListener callback) {
     @Override
     protected Integer doInBackground() {
         try {
-            // Call the List api method only if server call for list was not done before and return the unread count.
-            if (!ApplozicClient.getInstance(ApplozicService.getContextFromWeak(weakReferenceContext)).wasServerCallDoneBefore(null, null, null)) {
-                if (!Utils.isInternetAvailable(ApplozicService.getContextFromWeak(weakReferenceContext))) {
-                    return null;
+            String data = new MessageClientService(ApplozicService.getContextFromWeak(weakReferenceContext)).getMessages(null,null,null,null,null,false);
+            JsonObject jsonObject = JsonParser.parseString(data).getAsJsonObject();
```
pranaysaini04, line 35: name of the field is vague here, change it to something which is more aligned to its usability

himanshujoshi848: fixed

pranaysaini04 (approved):

PR #496, opened 2024-03-21
[CM-1495] Text Align for Bottom Navigation Fragment fix

himanshu-kt:
Fixed issue: Text Align for Bottom navigation Fragment | Android Agent app

image

### DIFF
--- a/kommunicate/src/main/java/io/kommunicate/utils/KmUtils.java
+++ b/kommunicate/src/main/java/io/kommunicate/utils/KmUtils.java
@@ -18,6 +18,7 @@ import android.text.TextUtils;
 import android.view.View;
 import android.view.Window;
 import android.view.WindowManager;
+import android.widget.ImageView;
 import android.widget.RelativeLayout;
 import android.widget.TextView;
 import android.widget.Toast;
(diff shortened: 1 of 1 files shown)


himanshujoshi848 (approved):

himanshujoshi848 (approved):

kommunicate/src/main/java/io/kommunicate/utils/KmUtils.java
```diff
@@ -77,6 +78,11 @@ public static void setDrawableTint(TextView textView, int colorId, int index) {
                 .setColorFilter(new PorterDuffColorFilter(colorId, PorterDuff.Mode.SRC_IN));
     }
 
+    public static void setDrawableTint(ImageView imageView, int colorId) {
```
Sathyan-KM, line 81: from where we are calling this function?

himanshu-kt: Earlier we were calling `setDrawableTint(Textview ...)` from agent app, but now its not called from anywhere.
We are now calling `setDrawableTint(ImageView ...)` from agent app. 

image

Same PR is raised in agent app too, since there is modification in both. https://github.com/Kommunicate-io/AgentApp/pull/321/files

Sathyan-KM (approved):

PR #501, opened 2024-04-06
[CM-1452] - Attachment options visibility customization

himanshu-kt:
Exposed a customization in `applozic-settings.json` as `"hideAttachmentOptionsBeforeHandover": false` so that Attachments Options can be hidden when bot handles the chat. Once handoff happened (human assignee) Attachment Options should be visible.

### DIFF
--- a/app/src/main/assets/applozic-settings.json
+++ b/app/src/main/assets/applozic-settings.json
@@ -182,5 +182,5 @@
   "isMultipleAttachmentSelectionEnabled": false,
   "isImageCompressionEnabled": false,
   "minimumCompressionThresholdForImagesInMB": 5,
-  "hideAttachmentOptionsBeforeHandover": false
+  "hideAttachmentOptionsWithBots": false
 }
--- a/app/src/main/assets/applozic-settings.json
+++ b/app/src/main/assets/applozic-settings.json
@@ -181,5 +181,6 @@
   "chatBarTopLineViewColor": "",
   "isMultipleAttachmentSelectionEnabled": false,
   "isImageCompressionEnabled": false,
-  "minimumCompressionThresholdForImagesInMB": 5
+  "minimumCompressionThresholdForImagesInMB": 5,
+  "hideAttachmentOptionsBeforeHandover": false
 }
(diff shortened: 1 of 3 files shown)


app/src/main/assets/applozic-settings.json
```diff
@@ -181,5 +181,6 @@
   "chatBarTopLineViewColor": "",
   "isMultipleAttachmentSelectionEnabled": false,
   "isImageCompressionEnabled": false,
-  "minimumCompressionThresholdForImagesInMB": 5
+  "minimumCompressionThresholdForImagesInMB": 5,
+  "hideAttachmentOptionsBeforeHandover": false
```
Sathyan-KM, line 185: change the flag to `hideattachmentoptionswithbots`

kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/conversation/fragment/MobiComConversationFragment.java
```diff
@@ -4451,6 +4462,21 @@ private void checkForCustomInput() {
...
+        }
+    }
+
+    private void updateAttachmentOptionsVisibilityAfterHandover() {
+        if (isHandoverHappened && alCustomizationSettings.getHideAttachmentOptionsBeforeHandover()) {
+            if (attachmentIconLayout != null) {
+                attachmentIconLayout.setVisibility(VISIBLE);
+            }
+        }
+    }
```
Sathyan-KM, line 4478: Change the naming with bot or bot assignee instead of handover. Did you test the case of reopening the conversation which is already assigned to human (after killing the app)

himanshu-kt: Changed the names accordingly and have test that case.

app/src/main/assets/applozic-settings.json
himanshu-kt: Okay

Sathyan-KM (approved):

Sathyan-KM (approved):

PR #506, opened 2024-04-25
[CM-2014] Action Buttons in List RM display fix

himanshu-kt:


kommunicateui/src/main/res/layout/km_rich_message_receiver_layout.xml
```diff
@@ -86,91 +86,24 @@
             android:layout_height="wrap_content"
             android:layout_gravity="center_horizontal" />
 
-        <View
```
himanshu-kt, line 89: Buttons for ListRM are earlier harcoded as three Text Views.

Sathyan-KM (approved):

PR #529, opened 2024-07-22
[CM-2128] Improve Tap Gesture for attachments | Android SDK 

AbhijeetRanjan308:
## Summary
- Added a tap gesture of mainLayout if the file is downloaded  
- The attachment bubble will work only when the file is downloaded. User needs to click on download button to download.

## Video

https://github.com/user-attachments/assets/327f9421-dded-4870-8d02-572f95e171f7

### DIFF
--- a/kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/attachmentview/KmDocumentView.java
+++ b/kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/attachmentview/KmDocumentView.java
@@ -332,7 +332,7 @@ public class KmDocumentView {
 
         mainLayout.setOnClickListener(new View.OnClickListener() {
             public void onClick(View v) {
-                if (!isDownloadRequire()) { return; } // The following code will not be executed if the file is already downloaded.
+                if (!isDownloadRequire()) { return; } // The following code will not be executed if the file is not downloaded.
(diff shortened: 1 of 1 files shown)


kommunicateui/src/main/java/com/applozic/mobicomkit/uiwidgets/attachmentview/KmDocumentView.java
```diff
@@ -330,6 +330,25 @@ public void onAction(boolean didGrant) {
             }
         });
 
+        mainLayout.setOnClickListener(new View.OnClickListener() {
+            public void onClick(View v) {
+                if (!isDownloadRequire()) { return; } // The following code will not be executed if the file is already downloaded.
+                if (kmStoragePermissionListener.isPermissionGranted()) {
+                    playAudio();
```
adarshmishra, line 337: Why the method name is playAudio? What exactly it does with attachments like PDF

AbhijeetRanjan308: All the attachment files and audio files are using the same function to preview the file or play the audio.

adarshmishra: As discussed change the name accordingly in future checkins

PR #530, opened 2024-07-22
CUS-229-constant

Pratham02Komm:
Removed the ambiguity of 2 variable with same name


Pratham02Komm:
had filed another PR with CUS-229-updated. Closing this redundant PR

PR #533, opened 2024-07-30
CM-2127 PreChat Labels not visible when Configured from dashboard | Android SDK

Pratham02Komm:
### Summary

- Earlier the issue was that when the user tried to build conversation with preChat, it used to show default Pre Chat Screen with empty fields
- The function (launchConversationWithPreChat) was not in use and this function (launchConversationWithPreChat) was calling an important function (checkForLeadCollection).
- So I made use of this function in the (createOrLaunchConversation) function in KmConversationHelper.java, where it performs creation as well as the launch of conversation.
- After that, it is working fine, and whenever a user will try to do setwithPreChat, then they will get desired results as attached below


AbhijeetRanjan308 (commented):
Any specific reason for removing try catch block.

AbhijeetRanjan308 (approved):

PR #545, opened 2024-09-17
Fix for AutoSuggestion not working| Android Agent App

AbhijeetRanjan308:
## Summary
- The data type for `CREATED_AT` and `UPDATED_AT` fields in the API response for Auto Suggestion (`rest/ws/autosuggest/message/`) has changed. Previously, these fields returned a Long type, but they are now returned as String.

### DIFF
--- a/kommunicate/src/main/java/io/kommunicate/models/KmAutoSuggestionModel.java
+++ b/kommunicate/src/main/java/io/kommunicate/models/KmAutoSuggestionModel.java
@@ -8,8 +8,8 @@ public class KmAutoSuggestionModel extends JsonMarker {
     private String category;
     private String content;
     private boolean deleted;
-    private long created_at;
-    private long updated_at;
+    private String created_at;
+    private String updated_at;
     private String userName;
     private long id;
     private boolean supportsRichMessage = false;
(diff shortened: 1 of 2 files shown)


kommunicate/src/main/java/io/kommunicate/models/KmAutoSuggestionModel.java
```diff
@@ -8,8 +8,8 @@ public class KmAutoSuggestionModel extends JsonMarker {
     private String category;
     private String content;
     private boolean deleted;
-    private long created_at;
-    private long updated_at;
+    private String created_at;
```
devashishmamgain, line 11: these are timestamp in UTC, need to find a way to store it as Date field so that anyplace if we have to use the date time, we can easily do that by converting to user's timezone

devashishmamgain (changes requested):

kommunicate/src/main/java/io/kommunicate/models/KmAutoSuggestionModel.java
devashishmamgain: same logic for iOS too

AbhijeetRanjan308: okay

AbhijeetRanjan308:
Closing PR as changes are done on server side to fix this issue.

PR #562, opened 2024-10-14
Updated Root Detection Logic

prateek-kommunicate:
Changed the way of checking root detection from normal Java approach to native CPP approach. Because from Frida Java approach can be bypassed easily. But with context of CPP approach it is difficult to bypass because it executes the process at lower level also Frida is build mostly for Java in android side so it can intercept and manuplate that code easily.

Fix Video:
<video src="https://github.com/user-attachments/assets/eb27dafc-08e7-49d5-880a-243366b50cb0" />


prateek-kommunicate:
As discussed closing this PR by merging it into the VPAT branch which will not go in **production**. i.e The code in this PR is not meant to be merge in development branch and release. It is for **VPAT** only

PR #620, opened 2025-04-09
Rename files and resources with applozic or al instance android sdk

prateek-kommunicate:


kommunicateui/src/main/java/io/kommunicate/ui/attachmentview/KmDocumentView.java
```diff
@@ -86,17 +86,17 @@ public KmDocumentView(Context context, KmStoragePermissionListener kmStoragePerm
...
-        fileText = (TextView) rootview.findViewById(R.id.applozic_doc_file_name);
-        uploadDownloadImage = (ImageView) rootview.findViewById(R.id.applozic_download_image);
+        progressBar = (ProgressBar) rootview.findViewById(R.id.doc_download_progress);
+        sizeTextView = (TextView) rootview.findViewById(R.id.doc_file_size);
+        fileText = (TextView) rootview.findViewById(R.id.doc_file_name);
+        uploadDownloadImage = (ImageView) rootview.findViewById(R.id.core_download_image);
         docIcon = (ImageView) rootview.findViewById(R.id.doc_icon);
         ImageView cancelIcon = (ImageView) rootview.findViewById(R.id.download_calcle_icon);
-        audioseekbar = (SeekBar) rootview.findViewById(R.id.applozic_audio_seekbar);
+        audioseekbar = (SeekBar) rootview.findViewById(R.id.audio_seekbar);
```
adarshmishra, line 99: @prateek-kommunicate these resources name are very generic and can conflict with app who is integrating our sdk. We had this issue in past. Pleas validate and add some prefix like core, km etc in prefix

prateek-kommunicate: This will not create confict as kommunicate has its own R file i.e it will he as written below. So, It wont collide with any other classes or file name present in SDK users app.

io.kommuicate.ui.R.id. audio_seekbar

adarshmishra (approved):

PR #624, opened 2025-04-22
Add exception handling for SQLite Cipher Database migration.

prateek-kommunicate:


kommunicate/src/main/java/io/kommunicate/devkit/database/MobiComDatabaseHelper.java
```diff
@@ -257,7 +259,12 @@ public MobiComDatabaseHelper(Context context, String name, SQLiteDatabase.Cursor
         super(context, name, factory, version);
         SQLiteDatabase.loadLibs(context);
         if (!DBUtils.isDatabaseEncrypted(context, name)) {
-            DatabaseMigrationHelper.migrateDatabase(context, name);
+            try {
+                DatabaseMigrationHelper.migrateDatabase(context, name);
+            } catch (Exception e) {
```
adarshmishra, line 264: What happens after caching the exception, looks like we are only logging it? Will app continue working without encrypted DB?

prateek-kommunicate: Yes, In that case the It will not migrate to sqlcipher and the old database will be used.

prateek-kommunicate: It will retry doing the same when the app is launched again.

adarshmishra: What is the chance retry will succeed in next app launch? It will flood our sentry. Better approach is either prompt user to logout/update and stop trying after few attempts.

adarshmishra (approved):

PR #653, opened 2026-03-09
Adding support for security token for images

adarshmishra:
- Token will be fetched from appSetting
- Will be added in rich message Images passed in payload

### DIFF
--- a/kommunicate/src/main/java/io/kommunicate/models/KmAppSettingModel.java
+++ b/kommunicate/src/main/java/io/kommunicate/models/KmAppSettingModel.java
@@ -147,6 +147,7 @@ public class KmAppSettingModel extends JsonMarker {
         private boolean disableChatWidget;
         private int csatRatingBase = 3;
         private int botTypingIndicatorInterval;
+        private String sasT;
 
         public String getPosition() {
             return position;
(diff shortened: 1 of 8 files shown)


kommunicate/src/main/java/io/kommunicate/utils/KmAppSettingPreferences.kt
```diff
@@ -144,20 +148,54 @@ object KmAppSettingPreferences {
...
+        }
+
+    @JvmStatic
+    fun appendSasToken(url: String?): String? {
+        val currentToken = sasToken
+        if (url == null || currentToken.isNullOrEmpty()) {
+            Log.d(TAG, "Not appending sasToken. URL is null: ${url == null}, Token is null/empty: ${currentToken.isNullOrEmpty()}")
+            return url
+        }
+        val finalUrl = if (url.contains("?")) {
```
devashishmamgain, line 191: it will go with some parameter name like sasToken=....
or just the token without parameter name?

kommunicate/src/main/java/io/kommunicate/models/KmAppSettingModel.java
```diff
@@ -147,6 +147,7 @@ public static class KmChatWidget extends JsonMarker {
         private boolean disableChatWidget;
         private int csatRatingBase = 3;
         private int botTypingIndicatorInterval;
+        private String sasT;
```
devashishmamgain, line 150: we can name it sasToken? if that's what T stands for here.

kommunicate/src/main/java/io/kommunicate/utils/KmAppSettingPreferences.kt
adarshmishra: whole thing will be stored on backend. Will be appended to image Url

```<imageurl>?<token>```

devashishmamgain (approved):

PR #684, opened 2026-09-07
TDH-4225 Prevent concurrent database migration failures

ayushsharmakm:
### Summary

Fixes a race condition where multiple SDK components could start database encryption migration at the same time.

### What changed

- Allowed only one database migration to run at a time.
- Rechecked the encryption status after getting the migration lock. If another thread already completed the migration, the waiting thread exits.
- Made `MobiComDatabaseHelper` initialization thread-safe using double-checked locking.
- Ensured database connections are always closed, even when migration fails.
- Deleted the temporary encrypted database and rethrew the error when data copying fails.

### Why this change was needed

Multiple threads could create separate database helper instances and start migration on the same database files. This could cause migration failures, incomplete temporary databases, or database corruption.

The updated flow ensures that only one thread initializes the helper and performs migration. Normal read/write calls return the existing instance without synchronization overhead. Migration errors are rethrown so the process stops before promoting an incomplete database; the caller catches and reports them to Sentry.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved database encryption migration reliability, including safer handling of interrupted or incomplete migrations.
  - Prevented unnecessary migration attempts for databases that are already encrypted.
  - Improved concurrent database initialization and migration behavior.
  - Added cleanup of temporary migration data to reduce the risk of stale or unusable database files.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved database encryption migration reliability, including safer temporary-file cleanup when migration fails.
  - Improved message synchronization by supporting JSON-encoded messages alongside existing message formats.
  - Added graceful handling for message deserialization failures, with automatic fallback to a full synchronization.
  - Improved database initialization reliability by preventing redundant encryption checks and reducing concurrency-related issues.


kommunicate/src/main/java/io/kommunicate/devkit/database/MobiComDatabaseHelper.java
```diff
@@ -274,7 +274,7 @@ public MobiComDatabaseHelper(Context context, String name, SQLiteDatabase.Cursor
         }
     }
 
-    public static MobiComDatabaseHelper getInstance(Context context) {
+    public static synchronized MobiComDatabaseHelper getInstance(Context context) {
```
adarshmishra: Is this called only at time of DB creation or called later as well? This will create performance issues if it is getting called everytime for read/write

kommunicate/src/main/java/io/kommunicate/database/DatabaseMigrationHelper.kt
```diff
@@ -49,28 +56,33 @@ object DatabaseMigrationHelper {
...
+                    copyDataBetweenDatabases(unencryptedDb, encryptedDb)
+                } finally {
+                    encryptedDb.close()
+                }
+            } finally {
+                unencryptedDb.close()
+            }
+        } catch (exception: Exception) {
+            android.database.sqlite.SQLiteDatabase.deleteDatabase(encryptedTempDbFile)
+            throw exception
```
adarshmishra, line 90: Do we need to throw exception?

kommunicate/src/main/java/io/kommunicate/devkit/database/MobiComDatabaseHelper.java
ayushsharmakm: It is called later as well. Updated to double-checked locking so synchronization occurs only during initial creation, subsequent calls return the existing instance without locking.

kommunicate/src/main/java/io/kommunicate/database/DatabaseMigrationHelper.kt
ayushsharmakm: Yes, we need to rethrow it. migrateDatabase() has only one caller the MobiComDatabaseHelper constructor which catches the exception and reports it to Sentry, so it won't cause an unhandled crash. Rethrowing stops the migration after cleaning up the temporary database. Otherwise, execution could continue to the promotion step with an incomplete or deleted temporary database, risking deletion of the original database.

adarshmishra (approved):

PR #687, opened 2026-09-15
TDH: 4225 Fix database migration replacement and handle missing message and thumbnail data

ayushsharmakm:
## Summary

This PR improves database migration safety and handles missing message-response and thumbnail data without SDK errors.

### 1. Atomically replace the database after migration

Previously, the original unencrypted database was deleted before renaming the encrypted database. If the rename failed, the application could be left without the original database.

The migration now:

- Uses `Os.rename()` to replace the database atomically.
- Keeps the original database safe if replacement fails.
- Throws a meaningful `SQLiteException` instead of only printing an error.

### 2. Handle missing message response data

The SDK called `.toString()` directly on the `message` and `userDetails` JSON fields. If either field was missing, it produced:

```text
NullPointerException: JsonElement.toString() on a null object reference
```

The SDK now:

- Checks `message` before processing it.
- Returns cached messages when `message` is missing.
- Handles missing `userDetails` safely.
- Marks the server call complete only after validating the message data.
- Applies the fix to both message-loading methods.

### 3. Skip nonexistent cached thumbnails

The SDK attempted to decode a cached thumbnail without checking whether the file existed, producing:

```text
BitmapFactory: Unable to decode file
FileNotFoundException: No such file or directory
```

The SDK now checks that the thumbnail file exists before decoding it. When missing, it uses the existing server-download fallback.

## Expected behavior

- Database migration cannot leave the application without its original database if replacement fails.
- Missing message data no longer produces a `NullPointerException`.
- Cached messages remain available when the message response is incomplete.
- Missing local thumbnails use the network fallback without generating a file-not-found decoder error.
- No logout, crash, or new user-facing toast is introduced.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved database migration reliability by ensuring replacements complete atomically and reporting migration failures consistently.
  * Fixed thumbnail caching so cached images are used only when the corresponding local file is actually available.
  * Improved conversation loading when message or user information is missing, preventing errors and preserving cached content where possible.
  * Enhanced network failure handling for invalid or incomplete conversation responses.


kommunicate/src/main/java/io/kommunicate/devkit/api/attachment/FileClientService.java
```diff
@@ -127,8 +127,9 @@ public Bitmap loadThumbnailImage(Context context, Message message, int reqWidth,
             options.inJustDecodeBounds = true;
             // Todo get the file format from server and append
             String imageName = FileUtils.getName(message.getFileMetas().getName()) + message.getCreatedAtTime() + "." + FileUtils.getFileFormat(message.getFileMetas().getName());
-            String imageLocalPath = getFilePath(imageName, context, message.getFileMetas().getContentType(), true).getAbsolutePath();
-            if (imageLocalPath != null) {
+            File thumbnailFile = getFilePath(imageName, context, message.getFileMetas().getContentType(), true);
+            String imageLocalPath = thumbnailFile.getAbsolutePath();
```
adarshmishra, line 131: Is it possible to thumbnailFile.getAbsolutePath() returns? Better to guard it

kommunicate/src/main/java/io/kommunicate/devkit/api/conversation/MobiComConversationService.java
```diff
@@ -385,20 +386,25 @@ public synchronized List<Message> getMessages(Function1<Message, Boolean> ignore
...
+            }
+
+            SettingsSharedPreference.getInstance(context).updateServerCallDoneStatus(contact, channel, conversationId);
+
             JSONObject jsonObject = new JSONObject(data);
             String channelFeedResponse = "";
             String conversationPxyResponse = "";
-            String element = parser.parse(data).getAsJsonObject().get("message").toString();
-            String userDetailsElement = parser.parse(data).getAsJsonObject().get("userDetails").toString();
+            JsonElement userDetailsElement = parser.parse(data).getAsJsonObject().get("userDetails");
```
adarshmishra: We are already paring jsonObject at like number 393, this we can reuse, parsing long string takes some execution

adarshmishra: Create JsonObject jsonObject = parser.parse(data).getAsJsonObject(); once and get message and userDetails wherever required

kommunicate/src/main/java/io/kommunicate/devkit/api/attachment/FileClientService.java
ayushsharmakm: getFilePath() always returns a File, and getAbsolutePath() does not return null. Therefore, the previous null check was redundant. If the file does not exist, BitmapFactory.decodeFile() returns null and the existing download fallback handles it.

kommunicate/src/main/java/io/kommunicate/devkit/api/conversation/MobiComConversationService.java
ayushsharmakm: Done. The JSON is now parsed once and the same JsonObject is reused for message, userDetails, and other required fields.

adarshmishra (approved):

PR #688, opened 2026-09-15
TDH-4265: Handle unreadable SQLCipher database without crashing

ayushsharmakm:
```markdown
## Error

The app crashed while opening an unreadable SQLCipher database:
```text
SQLiteNotADatabaseException: file is not a database (code 26)
```
The exception propagated through the background task and caused a fatal `RuntimeException`.

## What Changed

- Handled `SQLiteNotADatabaseException` in readable and writable database methods.
- Added `recoverUnreadableDatabase()` to recreate an unreadable database.
- Added synchronization to the database-opening methods.
- Stored the database name in the helper.
- Reset migration retries and sync timestamps during recovery.
- Added WAL checkpointing, `DELETE` journal mode, and a transaction to keep the source database consistent during migration.
- Added sidecar cleanup and atomic replacement of the original database with the encrypted copy.

## Why Changed

- `recoverUnreadableDatabase()` was added to close and delete the unreadable database, reset its state, and recreate it instead of crashing the app.
- `synchronized` was added to prevent multiple threads from deleting or recreating the same database concurrently.
- `databaseName` was stored so the recovery method can identify and delete the exact database file and its associated files.
- Sync timestamps are reset so conversations, messages, and metadata can be fetched again after recreating the empty database.
- Recovery is attempted only once. The original exception is rethrown if recovery cannot complete safely.
- Ensures all committed data is copied without concurrent database modifications.
- Prevents stale journal files or an incomplete replacement from leaving the database unreadable.

## Testing

Reproduced the issue by corrupting the AgentApp database. After the fix, the SDK recreated the database and the app continued running without a fatal exception.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved recovery when the app’s local database becomes unreadable.
  - The app now verifies required credentials before attempting database recovery.
  - When recovery is possible, the affected local database is recreated to restore access.
  - Original database errors are preserved when recovery cannot be completed.
  - Improved database migration reliability by ensuring pending changes are safely finalized before copying data.
  - Message retrieval now correctly reports unexpected server-response errors as network failures.


kommunicate/src/main/java/io/kommunicate/devkit/database/MobiComDatabaseHelper.java
```diff
@@ -291,14 +304,46 @@ public static MobiComDatabaseHelper getInstance(Context context) {
         return instance;
     }
 
-    public SQLiteDatabase getReadableDatabase() {
-        // The password is now handled by the constructor.
-        return super.getReadableDatabase();
+    public synchronized SQLiteDatabase getReadableDatabase() {
+        try {
+            return super.getReadableDatabase();
```
adarshmishra, line 308: have we tested the performance issues possibility due to synchronizing while getting database? Check if we can just synchronize recovery method?

ayushsharmakm: Tested during rapid navigation and pagination; warm calls completed in under 1 ms. SQLCipher 4.9.0 already synchronizes these getter methods internally. We kept the override synchronized to ensure corruption detection and recovery remain atomic, preventing multiple threads from starting recovery simultaneously.

adarshmishra: if we want only recovery to be synchronized, move it to recovery detection.

ayushsharmakm: Okay. I’ll move the synchronization to the recovery method and keep the normal database calls unchanged.

adarshmishra (approved):

